PR #30, opened 2019-07-30
Doxygen: Generate documentation from header files.

MatejKastak:
Currently the doxygen documentation is generated only from files in `src/` directory.
Files in `include/` have valid doxygen documentation but are not processed. Because of this classes like `RegexpUnit` are missing from generated documentation.

Retdec does something similar: https://github.com/avast/retdec/blob/master/doc/doxygen/doxygen.cfg.in#L828


MatejKastak:
Judging by the fact that `tests/` contains [`doxygen.h`](https://github.com/avast/yaramod/blob/master/tests/doxygen.h#L7)  I assume we also want to include this directory to full documentation.

PR #31, opened 2019-07-31
Python regular expression bindings

### LINKED ISSUE #29: Provide regexp units through Python bindings
Regular expressions are parsed down to individual regular expression units (character class, alternation, meta characters like `.`, `\s`, ...).  These unit are available through C++ API but they are not available through Python bindings. It would be nice to have them available in case anybody wants to process them from Python.

It should be enough to add all of them into `src/python/yaramod_python.cpp` with pointers to their respective C++ counterparts.

MatejKastak:
Implement new bindings for parsed regular expressions.

Added tests checking mentioned functionality.

This PR fixes #29.


tests/python/test_parser.py
```diff
@@ -933,3 +933,331 @@ def test_string_with_invalid_utf8_sequences(self):
...
+        self.assertEqual(len(rule.tags), 0)
+
+        regexp_string = rule.strings[0]
+        self.assertEqual(regexp_string.identifier, '$1')
+        self.assertTrue(regexp_string.is_regexp)
+        self.assertEqual(regexp_string.text, '/asd|1234/ ascii wide nocase fullword')
+        self.assertEqual(regexp_string.pure_text, b'asd|1234')
+        # Looks like the suffix modifiers are stored in multiple places
+        # and the only ones that are used are from String class.
+        # self.assertEqual(regexp_string.suffix_modifiers, '')
```
metthal: For what I mean with suffix modifiers, open https://perldoc.perl.org/perlre.html and scroll down to modifiers overview. YARA only supports `i` and `s` modifiers.

include/yaramod/types/regexp.h
```diff
@@ -365,6 +413,11 @@ class Regexp : public String
 		_suffixMods = suffixMods;
 	}
 
+	const std::shared_ptr<RegexpUnit>& getUnit() const { return _unit; }
+
+	void setUnit(const std::shared_ptr<RegexpUnit>& unit) { _unit = unit; }
+	void setUnit(std::shared_ptr<RegexpUnit>&& unit) { _unit = std::move(unit); }
```
s3rvac: I am just wondering, there are a lot of such overloaded functions (one taking an l-value ref-to-const, one r-value ref). Since they are in header files, wouldn't it be easier to use a single function taking a forwarding reference and using `std::forward()` instead?
```cpp
template <typename T>
void setUnit(T&& unit) { _unit = std::forward<T>(unit); }
```
Or do you prefer the current approach as it contains explicit types? Or forwarding references don't bode well with Python bindings?

This may not be directly related to the PR as there may be already other places with such overloads in the code, so I am also CCing @metthal.

metthal: Result of me not knowing this stuff like 4 years ago when I was writing very first version of yaramod. It should be sufficient to have single universal reference method and Python bindigs should cope with it but it'll require explicit specification of what type to infer. These interfaces will most probably change as we'll move towards unique pointers once we get rid of bison and flex so I don't necessarily require from Matej.

s3rvac: Thank you for the response! Consider it a tip to improve the code in the future then :slightly_smiling_face: I will mark the conversion as resolved as this is not directly related to the present PR.

metthal: We appreciate your tips. Will keep that in mind (cc @TadeasKucera).

PR #38, opened 2019-10-23
Parse with pog

TadeasKucera:
This PR brings [pog](https://github.com/metthal/pog) into yaramod to move on from flex and bison. We introduce _TokenStream_ concept which means that we remember all parsed tokens in one list that keeps their order. To avoid data inconsistance we keep all the data only in the _TokenStream_ and access the data by pointers (actually iterators). The main reason for bringing the _TokenStream_ was to parse and store comments in the input files in the exact positions they were written in.

This PR also introduces yaramod autoformatter of yara-rules: the _TokenStream_ class has _operator<<_ which does this trick, because it's output is formated in a proper manner. The whole process of parsing can be initiated from the _Yaramod_ class. To keep the data in _TokenStream_ only, we also had to modify the yaramod builders. User can now also supply their comments to the builders directly.


include/yaramod/builder/yara_expression_builder.h
```diff
@@ -47,9 +61,51 @@ class YaraExpressionBuilder
...
+	}
+	template<typename ExpPtr>
+	YaraExpressionBuilder(std::shared_ptr<TokenStream> ts, ExpPtr&& expr)
+		: _tokenStream( ts )
+		, _expr(std::forward<ExpPtr>(expr))
+	{
+	}
+	template<typename ExpPtr>
+	YaraExpressionBuilder(std::shared_ptr<TokenStream> ts, ExpPtr&& expr, const Expression::Type& type)
+		: _tokenStream( ts )
```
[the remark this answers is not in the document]
TadeasKucera: Thank you for noticing, let me fix it.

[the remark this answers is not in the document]
TadeasKucera: The `( ts )` is just wrong - thank you for noticing, let me fix it here.
Sometimes I use spaces inside parentheses for easier reading, though.

include/yaramod/builder/yara_hex_string_builder.h
```diff
@@ -110,20 +121,25 @@ YaraHexStringBuilder jumpRange(std::uint64_t low, std::uint64_t high);
...
 	std::vector<std::shared_ptr<HexString>> hexStrings;
-	return _alt(hexStrings, args...);
+	return _alt(ts, hexStrings, args...);
 }
 
 template <>
 YaraHexStringBuilder alt(const std::vector<YaraHexStringBuilder>& units);
 
-YaraHexStringBuilder _alt(std::vector<std::shared_ptr<HexString>>& hexStrings, const YaraHexStringBuilder& unit);
+YaraHexStringBuilder _alt(std::shared_ptr<TokenStream> ts, std::vector<std::shared_ptr<HexString>>& hexStrings, const YaraHexStringBuilder& unit);
```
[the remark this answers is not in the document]
TadeasKucera: Fixed it, thank you for pointing this out!

include/yaramod/builder/yara_rule_builder.h
```diff
@@ -26,6 +37,7 @@ class YaraRuleBuilder
 	/// @name Constructor
 	/// @{
 	YaraRuleBuilder();
+	YaraRuleBuilder(std::shared_ptr<TokenStream> tokenStream);
```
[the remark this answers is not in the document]
TadeasKucera: No, that is unnecessary, because accepting shared_ptr parameter by value immediately increases the shared_ptr counter by one only to initialize the local variable used in the constructer. Then another copy is created from this local variable. One copying is definitely enough so I fixed it everywhere I found this problem. Thank you!

src/python/yaramod_python.cpp
```diff
@@ -421,7 +426,9 @@ void addBuilderClasses(py::module& module)
 {
 	py::class_<YaraFileBuilder>(module, "YaraFileBuilder")
 		.def(py::init<>())
-		.def("get", &YaraFileBuilder::get, py::arg("recheck") = false)
+		.def("get", [](YaraFileBuilder& self, bool recheck) {
+				return self.get(recheck, nullptr);
+			}, py::arg("recheck") = true)
```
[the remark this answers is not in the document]
TadeasKucera: I changed it for some reason when I was testing the code and forgot to put it back. Fixed it now.

src/types/literal.cpp
```diff
@@ -4,111 +4,1152 @@
...
+
+const std::shared_ptr<Symbol>& Literal::getSymbol() const
+{
+	try
+	{
+		return std::get<std::shared_ptr<Symbol>>(_value);
+	}
+	catch (std::bad_variant_access& exp)
+	{
+		std::cerr << "Called getSymbol() of a TokenValue which holds " << *this << ". Index = " << _value.index() << std::endl << exp.what() << std::endl;
```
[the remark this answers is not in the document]
TadeasKucera: Alright, noted and fixed, thank you!

src/yaramod.cpp
```diff
@@ -8,48 +8,26 @@
...
+	_driver.setInput(filePath);
+	if(!_driver.isValid())
 		return nullptr;
-
 	std::unique_ptr<YaraFile> result;
-	if (driver.parse())
-		result = std::make_unique<YaraFile>(std::move(driver.getParsedFile()));
-
+	if(_driver.parse())
+		result = std::make_unique<YaraFile>(std::move(_driver.getParsedFile()));
```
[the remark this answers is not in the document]
TadeasKucera: As for now, this is necessary: Currently, the `ParserDriver::getParsedFile()` method behaves as getter and returns l-value reference to `_file` stored within. Either we can modify the interface and return r-value reference (and modify each test not to call 'getParsedFile()' repeatedly) or we can introduce new `ParserDriver::moveParsedFile()` method.

include/yaramod/parser/parser_driver.h
```diff
@@ -129,30 +333,50 @@ class ParserDriver
...
-	/// @name Method for handling anonymous strings
+	/// @name Methods for handling anonymous strings
 	/// @{
 	bool isAnonymousStringId(const std::string& stringId) const;
 	std::string generateAnonymousStringPseudoId();
 	/// @}
 
+	/// @name Methods for handling token streams
+	/// @{
+	std::shared_ptr<TokenStream> currentTokenStream() const { return _tokenStreams.top(); }
```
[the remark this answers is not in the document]
TadeasKucera: Yes, thank you, I already fixed it together with other fixes of ambiguous TokenStream copying.

include/yaramod/parser/parser_driver.h
```diff
@@ -129,30 +333,50 @@ class ParserDriver
...
+	void pushTokenStream(std::shared_ptr<TokenStream> ts) { _tokenStreams.push(ts); }
+	size_t currentTokenStreamCount() const { return _tokenStreams.size(); }
```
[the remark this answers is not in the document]
TadeasKucera: Fixed everywhere.

include/yaramod/types/expression.h
```diff
@@ -54,13 +65,37 @@ class Expression
...
+		_tokenStream = ts;
+	}
 	/// @}
 
 	/// @name Setter methods
 	/// @{
 	void setType(Expression::Type type) { _type = type; }
 	/// @}
 
+	// virtual void clear() = 0;
```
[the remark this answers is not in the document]
TadeasKucera: No, sorry, I have forgot to delete the comment.

include/yaramod/types/hex_string.h
```diff
@@ -70,17 +73,19 @@ class HexStringUnit
  */
 class HexString : public String
 {
+	friend YaraRuleBuilder;
```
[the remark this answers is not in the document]
TadeasKucera: Thank you for noticing this. The `String` class (base class of `HexString`) actually offers a getter of the attribute the `YaraRuleBuilder` was accessing. So I removed the `friend` and used the getter instead.

include/yaramod/types/hex_string.h
```diff
@@ -70,17 +73,19 @@ class HexStringUnit
...
 public:
 	/// @name Constructors
 	/// @{
-	explicit HexString(const std::vector<std::shared_ptr<HexStringUnit>>& units);
-	explicit HexString(std::vector<std::shared_ptr<HexStringUnit>>&& units);
+	explicit HexString(std::shared_ptr<TokenStream> ts, const std::vector<std::shared_ptr<HexStringUnit>>& units);
```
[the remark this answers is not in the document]
TadeasKucera: No, thank you, I fixed it earlier.

include/yaramod/types/hex_string.h
```diff
@@ -116,35 +121,37 @@ class HexStringNibble : public HexStringUnit
...
+		return output;
 	}
 	/// @}
 
 	virtual std::size_t getLength() const override { return 1; }
 	/// @}
 
 private:
-	std::uint8_t _value; ///< Value of the nibble
+	TokenIt _value; ///< Value of the nibble, int
```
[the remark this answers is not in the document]
TadeasKucera: It was supposed to indicate what type is stored in the variant inside `Token`. Here it is wrong, because the type is std::string in this case. Thank you for noticing.

include/yaramod/types/hex_string.h
```diff
@@ -192,15 +201,15 @@ class HexStringJump : public HexStringUnit
 		// If both low and high bound is defined and they are the same, it is the fixed jump.
 		if (_low.has_value() && _high.has_value() && _low.value() == _high.value())
 		{
-			ss << _low.value();
+			ss << _low.value()->getUInt64_t();
```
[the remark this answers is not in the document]
TadeasKucera: I have removed `isString` and similar methods from `Literal` class and replace these with single templated `Literal::is` function. Similarly, the templated `Literal::get()` has been introduced.

But I would prefer to keep the in the methods `getString`, 'getInt', etc. of `Token` class because it is more readable. Would that be alright?

include/yaramod/types/plain_string.h
```diff
@@ -25,19 +25,22 @@ class PlainString : public String
 public:
 	/// @name Constructors
 	/// @{
-	explicit PlainString(const std::string& text);
-	explicit PlainString(std::string&& text);
+	explicit PlainString(std::shared_ptr<TokenStream> ts, const std::string& text);
```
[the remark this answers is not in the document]
TadeasKucera: Not at all, removed.

src/types/literal.cpp
```diff
@@ -4,111 +4,1152 @@
...
+			return _formated_value.value();
+		else
+			return numToStr<double>( getDouble() );
+	}
+	else if (isSymbol())
+	{
+		assert(_formated_value);
+		return _formated_value.value();
+	}
+	std::cerr << "Unexpected index: '" << _value.index() << "'" << std::endl;
```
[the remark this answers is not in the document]
TadeasKucera: Removed all printouts, thank you for noticing!

include/yaramod/types/regexp.h
```diff
@@ -23,41 +26,79 @@ namespace yaramod {
...
 	virtual RegexpVisitResult accept(RegexpVisitor* v) override
 	{
 		return v->visit(this);
 	}
 
-	const std::string& getCharacters() const { return _characters; }
-	void setCharacters(const std::string& characters) { _characters = characters; }
+	void addCharacters(const std::string& text)
+	{
+		for( char c : text )
```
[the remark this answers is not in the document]
TadeasKucera: Thank you! Fixed everywhere.

include/yaramod/types/string.h
```diff
@@ -40,7 +43,37 @@ class String
...
+	explicit String(std::shared_ptr<TokenStream> ts, Type type, const std::string& id)
+		: _tokenStream(ts)
+		, _type(type)
+		, _mods(Modifiers::None)
+	{
+		assert(_tokenStream);
+		_id = _tokenStream->emplace_back(STRING_KEY, id);
+		_assign_token = _tokenStream->emplace_back(ASSIGN, "=");
+	}
+	explicit String(std::shared_ptr<TokenStream> ts, Type type, TokenIt id, TokenIt assign_token, uint32_t mods, std::vector<TokenIt> mods_strings)
```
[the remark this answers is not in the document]
TadeasKucera: Nope, removed.

src/builder/yara_expression_builder.cpp
```diff
@@ -343,9 +433,64 @@ YaraExpressionBuilder& YaraExpressionBuilder::operator>>(const YaraExpressionBui
...
+/**
+ * Puts comment in front of the expression.
+ *
+ * @param other The other expression.
+ *
+ * @return Builder.
+ */
+YaraExpressionBuilder& YaraExpressionBuilder::comment(const std::string& message, bool multiline, const std::string& indent)
+{
+	if(message != "")
```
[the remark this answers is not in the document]
TadeasKucera: Alright, I fixed it everywhere (245 occurences):
`if(` -> `if (`

src/builder/yara_expression_builder.cpp
```diff
@@ -839,9 +1145,54 @@ YaraExpressionBuilder conjunction(const std::vector<YaraExpressionBuilder>& term
  */
 YaraExpressionBuilder disjunction(const std::vector<YaraExpressionBuilder>& terms, bool linebreaks)
 {
+	for( const auto& bld : terms )
```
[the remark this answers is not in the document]
TadeasKucera: Fixed.

src/builder/yara_expression_builder.cpp
```diff
@@ -826,6 +1129,9 @@ YaraExpressionBuilder disjunction(const YaraExpressionBuilder& lhs, const YaraEx
  */
 YaraExpressionBuilder conjunction(const std::vector<YaraExpressionBuilder>& terms, bool linebreaks)
 {
+	for( const auto& bld : terms )
```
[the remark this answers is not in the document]
TadeasKucera: Fixed.

src/builder/yara_expression_builder.cpp
```diff
@@ -799,7 +1076,20 @@ YaraExpressionBuilder set(const std::vector<YaraExpressionBuilder>& elements)
  */
 YaraExpressionBuilder conjunction(const YaraExpressionBuilder& lhs, const YaraExpressionBuilder& rhs, bool linebreak)
 {
-	return YaraExpressionBuilder(std::make_shared<AndExpression>(lhs.get(), rhs.get(), linebreak));
+	if( !lhs.canBeBool() )
```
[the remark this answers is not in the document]
TadeasKucera: Fixed.

TadeasKucera:
> I went through the code, mostly code style but there are also some concerns I have about some parts of the code so please go through them and answer questions.

Thank you very much for your review. I went through the comments and fixed the code as described above in each reply. Some checks are still running, I believe we pass them soon.

src/yaramod.cpp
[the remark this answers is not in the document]
metthal: I'm fine with having `std::move` here then.

include/yaramod/builder/yara_rule_builder.h
```diff
@@ -8,11 +8,22 @@
 
 #include <string>
 
+#include <yaramod/yaramod_error.h>
```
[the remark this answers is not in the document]
metthal: Also keep them sorted in alphabetical order.

include/yaramod/parser/parser_driver.h
```diff
@@ -6,38 +6,239 @@
...
+		std::shared_ptr<String>,
+		std::shared_ptr<RegexpUnit>, //16
+		std::vector<std::shared_ptr<RegexpUnit>>,
+		TokenIt, //18
+		RegexpRangePair
+	>;
+
+	/// @name Constructors
+	/// @{
+	Value(const std::string& v) : _value(v) {}
```
metthal: Curious. Why can't there be single constructor like

```cpp
template <typename T>
Value(T&& v) : _value(std::forward<T>(v)) {}
```

?

include/yaramod/types/hex_string.h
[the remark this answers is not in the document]
metthal: Sure but why keeping `_t` suffix there. It's the internal name of C++ type `std::uint64_t`. At least call it `getUInt64` instead.

include/yaramod/types/regexp.h
```diff
@@ -67,18 +109,37 @@ class RegexpClass : public RegexpUnit
...
 
 	virtual RegexpVisitResult accept(RegexpVisitor* v) override
 	{
 		return v->visit(this);
 	}
 
+	void addCharacters(const std::string& text)
+	{
+		for (char c : text)
+			_characters.push_back(_tokenStream->emplace_back(TokenType::REGEXP_CHAR, std::string() + c));
```
metthal: What is this construct `std::string()`? What's the goal? To convert `c` to string? If so, then use constructor (2) https://en.cppreference.com/w/cpp/string/basic_string/basic_string

include/yaramod/types/regexp.h
```diff
@@ -23,41 +26,80 @@ namespace yaramod {
...
-	void setCharacters(const std::string& characters) { _characters = characters; }
+	void addCharacters(const std::string& text)
+	{
+		for (char c : text)
+			_characters.push_back(_tokenStream->emplace_back(TokenType::REGEXP_CHAR, std::string() + c));
+	}
+
+	std::string getCharacters() const
+	{
+		std::stringstream ss;
```
metthal: I feel like these `stringstream` operations will hold us back on performance. `stringstream` is extremely slow compare to for example [libfmt](https://github.com/fmtlib/fmt) which already comes bundled with `pog`. It would probably make sense to do performance testing one more time so we can see whether there are possibly some issues with string stream performance.

include/yaramod/types/regexp.h
```diff
@@ -257,25 +313,35 @@ class RegexpOperation : public RegexpUnit
...
 	void setOperand(std::shared_ptr<RegexpUnit>&& operand) { _operand = std::move(operand); }
 
 protected:
-	RegexpOperation(char operation, std::shared_ptr<RegexpUnit>&& operand, bool greedy) : _operation(operation), _operand(std::move(operand)), _greedy(greedy) {}
+	RegexpOperation(TokenType operation_token_type, char operation_symbol, std::shared_ptr<RegexpUnit>&& operand, bool greedy)
+		: _operand(std::move(operand))
+	{
+		//take the operand's tokenStream and append first the operation and then greedy
+		_tokenStream = std::move(_operand->getTokenStream());
+		_operation = _tokenStream->emplace_back(operation_token_type, std::string() + operation_symbol);
```
metthal: Same as before, if you are trying to construct string from char then use constructor (2) https://en.cppreference.com/w/cpp/string/basic_string/basic_string

include/yaramod/types/regexp.h
```diff
@@ -368,13 +454,22 @@ class RegexpRange : public RegexpOperation
...
 	}
 
-	const std::pair<std::optional<std::uint64_t>, std::optional<std::uint64_t>>& getRange() const
+	std::pair<std::optional<std::uint64_t>, std::optional<std::uint64_t>> getRange() const
 	{
-		return _range;
+		std::optional<std::uint64_t> out1;
+		std::optional<std::uint64_t> out2;
+		if (_first)
+			out1 = std::make_optional(_first.value()->getUInt64_t());
```
metthal: Is this `make_optional` necessary here?

include/yaramod/types/regexp.h
```diff
@@ -437,7 +548,12 @@ class RegexpGroup : public RegexpUnit
 class RegexpConcat : public RegexpUnit
 {
 public:
-	RegexpConcat(std::vector<std::shared_ptr<RegexpUnit>>&& units) : _units(std::move(units)) {}
+	RegexpConcat(std::vector<std::shared_ptr<RegexpUnit>>&& units)
+		: _units(std::move(units))
+	{
+		for (auto unit : _units)
```
metthal: Do `for (const auto& unit : _units)` or `for (auto&& unit : _units)`. You are making copies of shared pointers for each iteration right now.

include/yaramod/types/regexp.h
```diff
@@ -505,24 +648,52 @@ class Regexp : public String
 	*                    ^
 	* @endcode
 	*/
-	const std::string& getSuffixModifiers() const
+	std::string getSuffixModifiers() const
 	{
-		return _suffixMods;
+		if (!_suffixMods)
+			return "";
```
metthal: Use `std::string{}` for empty string literals.

include/yaramod/types/regexp.h
```diff
@@ -505,24 +648,52 @@ class Regexp : public String
...
+		else
+			return (*_suffixMods)->getPureText();
```
metthal: Also why can't you do `return _suffixMods.value_or(std::string{});`

src/types/rule.cpp
```diff
@@ -165,19 +207,12 @@ const Expression::Ptr& Rule::getCondition() const
...
-/**
- * Returns the tags of the YARA rule.
- *
- * @return Tags.
- */
-const std::vector<std::string>& Rule::getTags() const
+std::vector<std::string> Rule::getTags() const
 {
-	return _tags;
+	std::vector<std::string> output;
```
metthal: If you know the size of `std::vector` beforehand then reserve the memory for it so you don't perform more allocations than necessary.

src/types/rule.cpp
```diff
@@ -248,7 +283,17 @@ void Rule::setMetas(const std::vector<Meta>& metas)
...
 void Rule::setTags(const std::vector<std::string>& tags)
 {
-	_tags = tags;
+	TokenIt last;
+	//delete all tags from tokenStream
+	for (const TokenIt& it : _tags)
+		last = _tokenStream->erase(it);
+	_tags = std::vector<TokenIt>();
+	// Insert new tags into TokenStream
+	for (const std::string& tag : tags)
```
metthal: Why not `const auto&`?

include/yaramod/types/token.h
```diff
@@ -0,0 +1,460 @@
...
+/**
+ * Class representing literal. Literal can be either
+ * string, integer or boolean. This class can only bear
+ * one literal type at the same time. Behavior is undefined
+ * when other type than actual type of the literal is requested.
+ *
+ * Caution: Integral literals are stored as string to preserve base
+ * and all preceding zeroes.
+ */
+class Literal
```
metthal: Why can't `Literal` be in its own file?

include/yaramod/types/token.h
```diff
@@ -0,0 +1,460 @@
...
+	ONELINE_COMMENT,
+	COMMENT,
+	INCLUDE_DIRECTIVE,
+	INCLUDE_PATH,
+	FUNCTION_CALL_LP,
+	FUNCTION_CALL_RP,
+	INVALID,
+};
+
+class Location
```
metthal: Why can't `Location` can't be in its own file?

include/yaramod/types/token.h
```diff
@@ -0,0 +1,460 @@
...
+#include <string>
+#include <sstream>
+#include <variant>
+
+#include "yaramod/yaramod_error.h"
+
+namespace yaramod {
+
+class Token;
+using TokenIt = std::list< Token >::iterator;
```
metthal: Spacing around angle brackets.

include/yaramod/types/token.h
```diff
@@ -0,0 +1,460 @@
...
+	template <typename T>
+	bool is() const
+	{
+		return std::holds_alternative< T >(_value);
+	}
+
+	bool isIntegral() const;
+	/// @}
+
+	friend std::ostream& operator<<(std::ostream& os, const Literal& literal) {
```
metthal: Curly bracket placement.

include/yaramod/types/token.h
```diff
@@ -0,0 +1,460 @@
...
+				literal._value
+			);
+		return os;
+	}
+
+private:
+	/// For an integral literal x there are two options:
+	/// i.  x it is unformatted:	_formated_value is empty  AND  _value contains x
+	/// ii. x it is formatted:	  _formated_value contains x's string representation  AND  _value contains pure x
+	std::variant< std::string, bool, int, int64_t, uint64_t, double, std::shared_ptr<Symbol> > _value; ///< Value used for all literals:
```
metthal: Spacing.

include/yaramod/types/token.h
```diff
@@ -0,0 +1,460 @@
...
+				 _type == TokenType::RP_ENUMERATION ||
+				 _type == TokenType::HEX_JUMP_RIGHT_BRACKET ||
+				 _type == TokenType::REGEXP_END_SLASH ||
+				 _type == TokenType::HEX_END_BRACKET ||
+				 _type == TokenType::RP_WITH_SPACE_BEFORE ||
+				 _type == TokenType::RP_WITH_SPACES;
+	}
+	/// @}
+
+	friend std::ostream& operator<<(std::ostream& os, const Token& token) {
```
metthal: Curly bracket placement.

include/yaramod/types/tokenstream.h
```diff
@@ -0,0 +1,135 @@
+#pragma once
```
metthal: I would be for renaming this file to `token_stream.h` because we refer to it as `TokenStream` so there should be some spacing between token and stream.

include/yaramod/types/tokenstream.h
```diff
@@ -0,0 +1,135 @@
...
+	/// @}
+protected:
+	void getTextProcedure(PrintHelper& helper, std::stringstream* os, bool withIncludes, bool alignComments);
+	void autoformat();
+	void determineNewlineSectors();
+	void addMissingNewLines();
+
+	std::optional<TokenIt> predecessor(TokenIt it);
+private:
+	std::list< Token > _tokens; ///< All tokens off the rule
```
metthal: Spacing.

include/yaramod/utils/utils.h
```diff
@@ -43,14 +45,22 @@ bool startsWith(const std::string& str, const T& withWhat)
...
-std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false)
+std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false, bool toUpper = false)
 {
 	std::ostringstream os;
 	if (showbase)
 		os << format << std::showbase << num;
 	else
 		os << format << num;
-	return os.str();
+	if (!toUpper)
```
metthal: There is already stream manipulator for making this uppercase https://en.cppreference.com/w/cpp/io/manip/uppercase

src/builder/yara_expression_builder.cpp
```diff
@@ -25,37 +43,21 @@ YaraExpressionBuilder logicalFormula(std::vector<YaraExpressionBuilder> terms, c
...
 	auto formula = op(terms[0], terms[1]);
 	for (std::size_t i = 2; i < terms.size(); ++i)
-		formula = op(formula, terms[i]);
+	{
+		if (!terms[i].canBeBool())
+		{
+			const auto& expr = terms[i].get();
+			error_handle("Expected boolean, got '" + expr->getText() + "' of type " + expr->getTypeString());
+		}
+		if (i>=2)
```
metthal: Spacing.

src/builder/yara_rule_builder.cpp
```diff
@@ -85,7 +122,42 @@ YaraRuleBuilder& YaraRuleBuilder::withModifier(Rule::Modifier mod)
...
+	TokenIt insert_before = _mod.value_or(_rule_it);
+	std::stringstream ss;
+	if (multiline)
+	{
+		ss << "/* " << comment << " */";
+		_tokenStream->emplace(insert_before, TokenType::COMMENT, ss.str());
+	}
+	else
+	{
+		for (auto item : comment)
```
metthal: There is already `find` method (https://en.cppreference.com/w/cpp/string/basic_string/find)

src/builder/yara_rule_builder.cpp
```diff
@@ -99,7 +171,25 @@ YaraRuleBuilder& YaraRuleBuilder::withTag(const std::string& tag)
  */
 YaraRuleBuilder& YaraRuleBuilder::withStringMeta(const std::string& key, const std::string& value)
 {
-	_metas.emplace_back(key, Literal(value, Literal::Type::String));
+	if (key == "" || value == "")
```
metthal: `std::string{}` for empty string literals.

src/python/yaramod_python.cpp
```diff
@@ -149,13 +145,17 @@ void addBasicClasses(py::module& module)
 		.def_property("value", &Meta::getValue, &Meta::setValue);
 
 	py::class_<Literal>(module, "Literal")
-		.def(py::init<std::string, Literal::Type>())
 		.def(py::init<bool>())
-		.def_property_readonly("text", &Literal::getText)
+		.def_property_readonly("text", [](Literal& self) { return self.getText(); })
```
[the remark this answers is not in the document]
metthal: This still holds as a question.

src/types/rule.cpp
```diff
@@ -332,18 +386,40 @@ void Rule::removeString(const std::string& id)
...
  * Removes tag from the rule.
  *
  * @param tag Tag to remove.
+ * @return true iff there was corresponding tag and it was removed
  */
 void Rule::removeTags(const std::string& tag)
 {
-	auto new_end = std::remove(_tags.begin(), _tags.end(), tag);
-	_tags.erase(new_end, _tags.end());
+	for (auto it = _tags.begin(); it != _tags.end(); ++it)
```
metthal: I would put the body of this for-loop inside curly braces to make it obvious that this multi-line condition belongs to it. I am fan of for-loops without curly braces for their bodies only in case of one-line bodies.

src/types/rule.cpp
```diff
@@ -332,18 +386,40 @@ void Rule::removeString(const std::string& id)
...
+		if ((*it)->getText() == tag)
+		{
+			_tokenStream->erase(*it);
+			_tags.erase(it);
```
metthal: It is not safe to iterate over vector while erasing from it. Read the description of `erase()` method on cppreference.

> Invalidates iterators and references at or after the point of the erase, including the end() iterator. 

https://en.cppreference.com/w/cpp/container/vector/erase

src/types/token.cpp
```diff
@@ -0,0 +1,332 @@
...
+#include <sstream>
+#include <stack>
+
+#include "yaramod/types/token.h"
+#include "yaramod/types/tokenstream.h"
+#include "yaramod/utils/utils.h"
+
+namespace yaramod {
+
+#define TABULATOR_LENGTH 8
```
metthal: Is this define used anywhere?

src/types/tokenstream.cpp
```diff
@@ -0,0 +1,604 @@
...
+}
+
+TokenIt TokenStream::find(TokenType type, TokenIt from)
+{
+	return find(type, from, end());
+}
+
+TokenIt TokenStream::find(TokenType type, TokenIt from, TokenIt to)
+{
+	return std::find_if (
```
metthal: Spacing.

src/types/tokenstream.cpp
```diff
@@ -0,0 +1,604 @@
...
+#include <cassert>
+#include <sstream>
+#include <stack>
+
+#include "yaramod/types/tokenstream.h"
+#include "yaramod/utils/utils.h"
+
+namespace yaramod {
+
+#define TABULATOR_LENGTH 8
```
metthal: I would turn preprocessor define into constexpr const value.

src/python/yaramod_python.cpp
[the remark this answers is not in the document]
TadeasKucera: The `Literal::getText` method has now one additional parameter `bool pure` which is defaultly set to `false`. When called with `true`, this method returns pure text (unescaped and without quotes). I didn't want to write simmilar code twice, so `Literal::getPureText` just returns `Literal::getText(true)`.

For this reason, we cannot bind `text` property to this function by function pointer:

```
E           TypeError: (): incompatible function arguments. The following argument types are supported:
E               1. (arg0: yaramod.Literal, arg1: bool) -> str
```

src/parser/parser_driver.cpp
```diff
@@ -4,75 +4,1535 @@
...
+	});
+	// $multiline_comment end
+
+	// $str
+	// $str tokens are not delegated with return Value but stored in _strLiteral
+	_parser.token(R"(\")").states("@default").enter_state("$str").action( [&](std::string_view) -> Value {
+		_strLiteral.clear();
+		return {};
+	});
+	_parser.token(R"(\\t)").states("$str").action( [&](std::string_view) -> Value {
```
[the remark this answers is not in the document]
TadeasKucera: I am not sure if we want this. We currently check escaped sequences and throw an exception when unknown escaped sequence occures, see test `ParserTest.InvalidEscapeSequence`, where we expect the parser to throw an exception when given input containing string `\n\r`.

src/parser/parser_driver.cpp
```diff
@@ -4,75 +4,1535 @@
...
+		_strLiteral += "\\t";
+		return {};
+	});
+	_parser.token(R"(\\n)").states("$str").action( [&](std::string_view) -> Value {
```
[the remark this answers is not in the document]
TadeasKucera: Same as above answer.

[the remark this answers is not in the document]
TadeasKucera: Again not sure if we want to lose unknown escaped sequence check.

include/yaramod/builder/yara_expression_builder.h
```diff
@@ -34,6 +36,19 @@ enum class IntFunctionEndianness
...
 };
 
+/**
+ * Represents error during parsing.
+ */
+class YaraExpressionBuilderError : public YaramodError
+{
+public:
+	YaraExpressionBuilderError(const std::string& errorMsg)
+		: YaramodError(std::string("YaraExpressionBuilder error: ") + errorMsg)
```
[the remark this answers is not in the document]
TadeasKucera: Fixed it, thank you!

include/yaramod/parser/parser_driver.h
TadeasKucera: Actually, there can be a single constructor like 
```
template <typename T>
Value(T&& v) : _value(std::forward<T>(v)) {}
```
I fixed it.

include/yaramod/types/regexp.h
TadeasKucera: Yes, that was my intention. I used the second constructor instead, thank you.

include/yaramod/types/regexp.h
TadeasKucera: Fixed.

include/yaramod/types/regexp.h
TadeasKucera: No, `std::optional<T>::operator=` can take `T` directly.

include/yaramod/types/regexp.h
TadeasKucera: Fixed everywhere.

include/yaramod/types/regexp.h
TadeasKucera: Because the types differ: `_suffixMods`'s type is `std::optional<TokenIt>` and I need to access the `Token::getPureText()` method.

src/types/rule.cpp
TadeasKucera: Alright, fixed

include/yaramod/types/tokenstream.h
TadeasKucera: Done

src/types/rule.cpp
TadeasKucera: Ok, I used `std::find_if` to find the iterator and then use it to remove the tag.

src/types/token.cpp
TadeasKucera: It is used in the `token_stream.cpp` and `token.cpp` files.

include/yaramod/types/hex_string.h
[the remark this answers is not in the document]
TadeasKucera: That's a good point! I renamed the `getInt64_t`, `getUInt64_t`, `isInt64_t` and `isUInt64_t` and removed the `_t`.

src/parser/parser_driver.cpp
```diff
@@ -4,75 +4,1535 @@
...
+		_strLiteral += "\\n";
+		currentLocation().addLine();
+		return {};
+	});
+	_parser.token(R"(\\x[0-9a-fA-F]{2})").states("$str").action( [&](std::string_view str) -> Value {
```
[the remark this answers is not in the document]
TadeasKucera: Again not sure if we want to lose unknown escaped sequence check.

include/yaramod/types/token.h
TadeasKucera: I moved it into separate file `location.h` (and `location.cpp`).

include/yaramod/types/token.h
TadeasKucera: I moved it into separate file `literal.h` (and `literal.cpp`).

src/parser/parser_driver.cpp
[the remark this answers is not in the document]
metthal: Good point but I still don't think we need special tokenizer actions for each of these sequences. There can be just single tokenizer action for all valid ones.

src/types/tokenstream.cpp
TadeasKucera: That is a very good point, thank you. Fortunately, the macro didn't cause any memory errors, but it will be much better to use `constexpr` instead.

src/parser/parser_driver.cpp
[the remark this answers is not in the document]
TadeasKucera: You are right, I eliminated the number of token definitions. Thank you.

src/parser/parser_driver.cpp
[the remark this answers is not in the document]
TadeasKucera: Done.

src/parser/parser_driver.cpp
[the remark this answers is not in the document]
TadeasKucera: Done.

include/yaramod/types/regexp.h
TadeasKucera: I have done performance testing on our cpp test suite builded with `RelWithDebInfo` build type. I searched for `stringstream` in the output in KCacheGrind and as you can see in the enclosed screenshot, there is not much time spent on `stringstream` operations.
Screenshot from 2019-11-15 14-38-13

PR #44, opened 2019-12-04
Fixes of old errors present already in Flex and Bison parser

TadeasKucera:
- construction of ArrayAccessEspression in parser
 - make sure IdExpression::_symbol is never null
 - add a test for IdExpression architecture


include/yaramod/types/expressions.h
```diff
@@ -1355,7 +1333,7 @@ class FunctionCallExpression : public IdExpression
 public:
 	template <typename ExpPtr, typename ExpPtrVector>
 	FunctionCallExpression(ExpPtr&& func, TokenIt left_bracket, ExpPtrVector&& args, TokenIt right_bracket)
-		: IdExpression()
+		: IdExpression(std::static_pointer_cast<const IdExpression>(func)->getSymbol())
```
TadeasKucera: Do we mind this cast here? I wanted to preserve the ExpPtr template for perfect forwarding...

metthal: It should be fine if used properly. And since it only should be created through builder then I consider it completely fine.

TadeasKucera: Perfect. It is used in the parser too, but carefully. It is only passed an expression that is actually IdExpression or derived class. (The "identifier" grammar rule always creates some IdExpression...)

tests/cpp/parser_tests.cpp
```diff
@@ -2428,22 +2428,65 @@ rule math_module {
...
+	EXPECT_TRUE(condition->isBool());
+	auto expEq = std::static_pointer_cast<const EqExpression>(condition);
+	EXPECT_EQ(R"(pe.version_info["CompanyName"] == "company")", expEq->getText());
+
+	auto expLeft = std::static_pointer_cast<const ArrayAccessExpression>(expEq->getLeftOperand());
+	EXPECT_EQ(R"(pe.version_info["CompanyName"])", expLeft->getText());
+	auto expItem = expLeft->getAccessor();
+	EXPECT_EQ(R"("CompanyName")", expItem->getText());
+	auto arraySymbol = expLeft->getSymbol();
+	EXPECT_NE(arraySymbol, nullptr);
```
TadeasKucera: Before this test, we didn't check if the constructed expression tree has proper structure. We only tested the getText method.

deps/pog/deps/fmt/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ ExternalProject_Add(
 	SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/fmt"
 	CMAKE_ARGS
 		-DCMAKE_INSTALL_PREFIX=${FMT_INSTALL_DIR}
-		-DCMAKE_BUILD_TYPE=Release
+		-DCMAKE_BUILD_TYPE=$<IF:$<CONFIG:Debug>,Debug,Release>
```
metthal: What is the motivation for using generator expression and not just `${CMAKE_BUILD_TYPE}` directly?

TadeasKucera: That is of course much better, thank you!

Issue #48, opened 2019-12-13
Autoformatting of file with CRLF line endings and comments results in file with mixed line endings


metthal:
If you take a file with CRLF line endings and put some comments in there, you end up with CR at the end of each comment while all other lines in the file have LF line endings. We should take care of it so that we end up with either only LF line endings (Linux + macOS) or CRLF line endings (Windows).

metthal:
Regarding the last point that we should have LF on Linux + macOS and CRLF on Windows, this would break a lot of things when one person wants to autoformat from Linux machine and the other one from Windows machine. They would rewrite their line endings each time.

The way it works should be probably that when file has CRLF line endings, it autoformats with CRLF but when file has LF line endings, it should autoformat with LF. This will also work hand in hand with git `autocrlf` option which causes git on Windows to checkout files with CRLF but turns them to LF before they are checked in into commit.

s3rvac:
Fixed in #49.

PR #49, opened 2019-12-16
Autoformat adds \r or \n depending on newlines used in the file.

TadeasKucera:
Bugfix of issue #48: We insert LF or CR or CRLF depending on what is used in the file, more precisely which of the three line endings `\n`, `\r` or `\r\n` has been used last time.

### DIFF
--- a/include/yaramod/types/expressions.h
+++ b/include/yaramod/types/expressions.h
@@ -1194,8 +1194,9 @@ public:
 	{
 	}
 
-	IdExpression(TokenIt symbol)
-		: _symbol(symbol->getSymbol())
+	IdExpression(TokenIt symbolToken)
+		: _symbol(symbolToken->getSymbol())
+		, _symbolToken(symbolToken)
 	{
 	}
 
@@ -1215,8 +1216,16 @@ public:
 		return _symbol;
 	}
 
+	void setSymbol(const std::shared_ptr<Symbol>& symbol)
+	{
+		_symbol = symbol;
+		if (_symbolToken)
+			_symbolToken.value()->setValue(_symbol, _symbol->getName());
+	}
+
 protected:
 	std::shared_ptr<Symbol> _symbol; ///< Symbol of the identifier
+	std::optional<TokenIt> _symbolToken; ///< Token of the 
 };
 
 /**
--- a/include/yaramod/types/token_stream.h
+++ b/include/yaramod/types/token_stream.h
@@ -125,8 +125,8 @@ public:
 
 	/// @name New Line Characters
 	/// @{
-	char getNewLineChar() const { return new_line_char; }
-	void setNewLineChar(char c) { new_line_char = c; }
+	const std::string& getNewLineStyle() const { return new_line_style; }
+	void setNewLineChar(std::string line) { new_line_style = std::move(line); }
 	/// @}
 
 	/// @name Reseting method
@@ -142,7 +142,7 @@ protected:
 private:
 	std::list<Token> _tokens; ///< All tokens off the rule
 	bool formatted = false; ///< The flag is set once autoformat has been called
-	char new_line_char = '\n'; ///< The character used for line endings: usually '\n' on Unix or '\r' on Windows
+	std::string new_line_style = "\n"; ///< The character used for line endings: usually '\n' on Unix or '\r' on MacOs or '\r\n' on Windows
 };
 
 } //namespace yaramod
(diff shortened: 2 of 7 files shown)


include/yaramod/types/token_stream.h
```diff
@@ -136,6 +142,7 @@ class TokenStream
 private:
 	std::list<Token> _tokens; ///< All tokens off the rule
 	bool formatted = false; ///< The flag is set once autoformat has been called
+	char new_line_char = '\n'; ///< The character used for line endings: usually '\n' on Unix or '\r' on Windows
```
s3rvac: Shouldn't it be `\r\n` on Windows instead of just `\r`?

src/parser/parser_driver.cpp
```diff
@@ -36,13 +36,14 @@ void ParserDriver::defineTokens()
 		currentLocation().addColumn(str.length());
 	});
 
-	_parser.token("\n").action([&](std::string_view str) -> Value {
+	_parser.token("[\n\r]").action([&](std::string_view str) -> Value {
+		currentTokenStream()->setNewLineChar(std::string{str}.front());
```
s3rvac: Tip: `std::string_view` has also [`front()`](https://en.cppreference.com/w/cpp/string/basic_string_view/front), so there is no need to convert the string view to a string.

include/yaramod/types/token_stream.h
TadeasKucera: So we will probably have to parse `\r\n` as a single character, not two different new line characters. Thank you for pointing this out.

TadeasKucera: I fixed the problem. We insert LF or CR or CRLF depending on what is used in the file, more precisely which of the three line endings `\n`, `\r` or `\r\n` has been used last time.

include/yaramod/types/token_stream.h
```diff
@@ -123,6 +123,12 @@ class TokenStream
 	std::vector<std::string> getTokensAsText() const;
 	/// @}
 
+	/// @name New Line Characters
+	/// @{
+	const std::string& getNewLineStyle() const { return new_line_style; }
+	void setNewLineChar(std::string&& line) { new_line_style = line; }
```
s3rvac: Beware that this copies `line` into `new_line_style`, i.e. it calls the copy-assignment operator. If you want it to get moved, you need to `std::move(line)` there:
```diff
-void setNewLineChar(std::string&& line) { new_line_style = line; }
+void setNewLineChar(std::string&& line) { new_line_style = std::move(line); }
```
The reason is that even though `line` is passed as an r-value to `setNewLineChar()`, inside the body, it is actually an l-value (it has a name: `line`). Thus, to make it an r-value, we need `std::move()`.

To simplify the implementation (and stop forcing callers of the function to pass an r-value), consider passing the parameter simply by value and move it into `new_line_style`:
```diff
-void setNewLineChar(std::string&& line) { new_line_style = line; }
+void setNewLineChar(std::string line) { new_line_style = std::move(line); }
```

include/yaramod/types/token_stream.h
s3rvac: Great :+1: BTW, if you want to simplify the implementation a bit, I think you can safely ignore `\r` because it is no longer used (it was used in older macOS versions) and consider just `\n` (Linux, macOS) and `\r\n` (Windows). What do you think @metthal?

include/yaramod/types/token_stream.h
TadeasKucera: Thank you very much! The second option seems to be better here, so I will pass by value.

include/yaramod/types/token_stream.h
TadeasKucera: I asked @metthal and he agrees with you so I will definitely simplify that: We will ignore `\r` because it is no longer used.

include/yaramod/types/token_stream.h
```diff
@@ -136,6 +142,7 @@ class TokenStream
 private:
 	std::list<Token> _tokens; ///< All tokens off the rule
 	bool formatted = false; ///< The flag is set once autoformat has been called
+	std::string new_line_style = "\n"; ///< The character used for line endings: usually '\n' on Unix or '\r' on MacOs or '\r\n' on Windows
```
metthal: Since these are private attributes, I think they should be prefixed with `_` to match conventions used everywhere throughout `yaramod`.

include/yaramod/types/expressions.h
```diff
@@ -1215,8 +1216,16 @@ class IdExpression : public Expression
 		return _symbol;
 	}
 
+	void setSymbol(const std::shared_ptr<Symbol>& symbol)
+	{
+		_symbol = symbol;
+		if(_symbolToken)
```
metthal: Space between if and `(` is missing.

Issue #46, opened 2019-12-11
Add switch for turning on/off Avast and VirusTotal specific symbols


metthal:
There are some Avast and VirusTotal specific symbols in symbol tables which would be nice to turn off using some switch when parsing the ruleset.

PR #51, opened 2020-01-08
Synchronise modules (issue #46)

TadeasKucera:
**Synchronise symbols with yara downstream and enable use of symbols from yara upstream only:**

- We have compared current yaramod modules with modules in Avast yara downstream and we make it consistent in this PR.

- We also change yaramod::ParserDriver to be able to avoid Avast-specific or VirusTotal-specific symbols when importing. Hence #46 Is solved. This will be benefitial for people outside Avast that need only symbols present in yara [upstream](https://github.com/VirusTotal/yara/tree/master/libyara/modules).


include/yaramod/parser/parser_driver.h
```diff
@@ -253,9 +253,9 @@ class ParserDriver
 	/// @name Constructors
 	/// @{
 	ParserDriver() = delete;
-  	ParserDriver(ParserMode parserMode);
-	explicit ParserDriver(const std::string& filePath, ParserMode parserMode = ParserMode::Regular);
-	explicit ParserDriver(std::istream& input, ParserMode parserMode = ParserMode::Regular);
+  	ParserDriver(ParserMode parserMode, bool avastSpecific = true, bool vtSpecific = true);
```
metthal: It would be best to avoid just boolean parameters if it's unclear from the context of the function name of what they suppose to mean. Calls to such functions/methods then look like this `ParserDriver(..., true, false)`. What exactly those `true` and `false` mean? The best way to do it would be through bitmasks because then you can do something like `ParserDriver(..., Features::VirusTotal | Features::Avast)`.

src/types/modules/module.cpp
```diff
@@ -78,22 +78,32 @@ bool Module::isInitialized() const
...
  *
  * @param name Name of the module to load
+ * @param avastSpecific Determines which symbols to import
  *
  * @return Module if found, @c nullptr otherwise.
  */
-std::shared_ptr<Module> Module::load(const std::string& name)
+std::shared_ptr<Module> Module::load(const std::string& name, bool avastSpecific)
 {
+	if (!avastSpecific && (name == "androguard" || name == "phish")) // the androguard and phish modules are completely avast-specific
```
metthal: I would like this to be an attribute of the module itself, like _which features you require in order to load_? This feels like a hack because information about a module is taken out from it.

src/types/modules/time_module.cpp
```diff
@@ -22,9 +22,10 @@ TimeModule::TimeModule() : Module("time")
  *
  * @return @c true if success, otherwise @c false.
  */
-bool TimeModule::initialize()
+bool TimeModule::initialize(bool avastSpecific)
 {
 	using Type = Expression::Type;
+	(void) avastSpecific;
```
metthal: What you can also do in order to avoid unused variable warnings is that you can specify parameter as `bool/* avastSpecific*/`). Basically, don't give a name to the parameter.

tests/cpp/parser_tests.cpp
```diff
@@ -2299,6 +2341,110 @@ rule same_variable_in_nested_for_loops {
...
+R"(
+import "androguard"
+
+rule dummy_rule {
+	condition:
+		androguard.max_sdk > androguard.signature.hits("dummy")
+}
+)");
+
+	Module::reset("androguard");
```
metthal: Will user need to call `Module::reset()` manually if they want to create more instances of yaramod? That doesn't sound very user friendly. This needs to happen all under the hood without user being aware of that. Please, change it so that user won't need to do this.

TadeasKucera: The user does not need to call the `reset` method, we only need it in our tests where we construct more ParserDriver instances - all sharing the same modules, because modules are stored as global variables in `module.cpp`. These modules need to be reseted if we need ParserDriver with different `Features`. This is something that the user will never need, because they will use the same `Features` among their `ParserDrivers`. Maybe we could stop storing modules as global variables but store them either inside `Yaramod` or inside `ParserDriver` classes.

metthal: But what if I want to create `Yaramod` instance once with VirusTotal features and once with no additional features enabled? Would I need to do that?

TadeasKucera: No, that would not require you calling the `reset` method. However, if you decided to use two different `Yaramod` instances, first one with `ImportFeature::Avast` and then second one with `ImportFeature::Basic` then yes, you would have to call the static method `Module::reset(<module_name>)`. The problem is, that both `ParserDriver` instances share the same modules. In my opinien, it is a good thing that the modules are shared between the instances. The whole mean of issue #46 was to reduce the number of symbols stored and the complexity of modules. Therefore I propose a solution which would keep it:

Along with the globally stored modules we will store also global `Feature` instance named `modules_state`. Whenever new instance of `Yaramod` with parameter `ImportFeature feature` is created, we look at `modules_state` and update it:
`modules_state |= feature`
possibly enabling new modules/symbols. With this approach, when loading a module, we will only look at the global variable `modules_state`.

What do you think?

metthal: > However, if you decided to use two different Yaramod instances, first one with ImportFeature::Avast and then second one with ImportFeature::Basic then yes, you would have to call the static method Module::reset(<module_name>). 

This is what I meant and I don't personally like it because it is so easy to misuse from user standpoint.

>  In my opinien, it is a good thing that the modules are shared between the instances.

Why do you think it is a good idea to keep the modules shared? I personally think that it was good idea when modules were immutable but I don't see any benefit when modules are possibly mutable.

metthal: I would propose to move from global modules to modules that are only local to certain `ParserDriver` instance. All the modules will have to be initialized multiple times when creating multiple `Yaramod` instances but I think that benefits of this solution outweigh benefits of having global modules.

TadeasKucera: I thought about it and I agree with you. We will get rid of global variables, we will get rid of the `reset` function or some hacks I proposed. It will be much cleaner and use just a liitle bit more memory to store the modules for each instance independently. I will implement it soon. Thank you for this discussion.

include/yaramod/parser/parser_driver.h
TadeasKucera: OK, I implemented this in the last commit.

TadeasKucera: OK, I implemented this in the last [commit](https://github.com/avast/yaramod/pull/51/commits/591e8499aeced53445819fa2b3602219cb4645a5).

src/types/modules/module.cpp
TadeasKucera: Understood, instead of specifying here names of unsupported modules we get this information out of the module itself. Implemented [here](https://github.com/avast/yaramod/pull/51/commits/591e8499aeced53445819fa2b3602219cb4645a5).

tests/cpp/parser_tests.cpp
TadeasKucera: Done: In [commit](https://github.com/avast/yaramod/pull/51/commits/ef2a4f3fb3eb01b73404fcc84284b5ec751cc60d) you will find a `ModulesPool` class which stores the modules. Each `ParserDriver` or `YaraFileBuilder` is assigned one instance of the `ModulesPool` class, which holds all used modules.

TadeasKucera: Done: In [commit](https://github.com/avast/yaramod/pull/51/commits/7afaecccd237d1cbc8e159bd8217962d7d3009c3) we use new class `ModulesPool` which stores the modules. Each `ParserDriver` or `YaraFileBuilder` instance is assigned its own instance of `ModulesPool` class.

metthal:
Thank you very much for the changes. Let's wait for the builds to finish and we'll merge it.

TadeasKucera:
Thank you for your suggestions! I believe we are ready for merge.

PR #57, opened 2020-01-22
Various fixes in parser and builder which were crashing yaramod or producing wrong rules

metthal:
* Builders now work properly when you create `YaraExpressionBuilder` out of already existing expression.
* Calculation of rule locations now works again.
* Include files are now closed as soon as possible to not exhaust file descriptors.
* Very last rule in the parsed file is now reported to be located in the correct file.
* Include guarded parsing mode now works properly again.


TadeasKucera:
Thank you Marek for all the fixes, I am very sorry there was so many problems.

Issue #58, opened 2020-01-28
Builder does not allow to create string meta with empty string


metthal:
`YaraRuleBuilder` does not allow to create string meta with empty string as value but it should allow it because even empty value is valid value. Exception thrown:

```
C++ exception with description "YaraRuleBuilder error: Error: String-Meta key and value must be non-empty." thrown in the test body.
```

PR #59, opened 2020-01-28
YaraRuleBuilder allows empty string meta values (issue #58)

TadeasKucera:
YaraRuleBuilder allows now to create string meta with empty string as value.

### DIFF
--- a/src/builder/yara_rule_builder.cpp
+++ b/src/builder/yara_rule_builder.cpp
@@ -176,8 +176,8 @@ YaraRuleBuilder& YaraRuleBuilder::withComment(const std::string& comment, bool m
  */
 YaraRuleBuilder& YaraRuleBuilder::withStringMeta(const std::string& key, const std::string& value)
 {
-	if (key == std::string{} || value == std::string{})
-		throw RuleBuilderError("Error: String-Meta key and value must be non-empty.");
+	if (key == std::string{})
+		throw RuleBuilderError("Error: String-Meta key must be non-empty.");
 
 	TokenIt insert_before = _strings_it.value_or(_condition_it);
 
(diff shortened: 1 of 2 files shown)


src/builder/yara_rule_builder.cpp
```diff
@@ -176,8 +176,8 @@ YaraRuleBuilder& YaraRuleBuilder::withComment(const std::string& comment, bool m
  */
 YaraRuleBuilder& YaraRuleBuilder::withStringMeta(const std::string& key, const std::string& value)
 {
-	if (key == std::string{} || value == std::string{})
-		throw RuleBuilderError("Error: String-Meta key and value must be non-empty.");
+	if (key == std::string{})
```
s3rvac: Tip: Here, we basically ask if `key` is empty. The more straightforward way would be to use `if (key.empty())`, which does exactly what we need and sounds better when read in English (_"if key is empty"_). Also, we will no longer need to create an empty string just to use it for comparison and then destroy it.

If you like the tip, feel free to commit it separately as it is not related to the present PR. I just wanted to point that out.

tests/cpp/builder_tests.cpp
```diff
@@ -141,6 +141,37 @@ RuleWithMetasWorks) {
 )", yaraFile->getTextFormatted());
 }
 
+TEST_F(BuilderTests,
+RuleWithEmptyStringMetasWorks) {
```
s3rvac: Consider naming the test `RuleWithEmptyStringMetaValueWorks` as we actually check that an empty _value_ is permitted (an empty key is still forbidden).

src/builder/yara_rule_builder.cpp
TadeasKucera: Thank you for the tip, I totally agree. Even though it is not related to this PR, I will include it because it is a small change of code.

tests/cpp/builder_tests.cpp
TadeasKucera: Thank you for your review, I have fixed the code as you suggested. The first of your tips will come as a separate PR.

PR #60, opened 2020-01-29
Use std::string.empty instead of comparing to empty string

TadeasKucera:
Use std::string.empty instead of comparing string instances to empty string

### DIFF
--- a/src/builder/yara_rule_builder.cpp
+++ b/src/builder/yara_rule_builder.cpp
@@ -90,7 +90,7 @@ std::unique_ptr<Rule> YaraRuleBuilder::get()
  */
 YaraRuleBuilder& YaraRuleBuilder::withName(const std::string& name)
 {
-	if (name == std::string{})
+	if (name.empty())
 		throw RuleBuilderError("Error: name must be non-empty.");
 	_name_it->setValue(name);
 	return *this;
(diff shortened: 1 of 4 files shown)


src/builder/yara_rule_builder.cpp
```diff
@@ -176,8 +176,8 @@ YaraRuleBuilder& YaraRuleBuilder::withComment(const std::string& comment, bool m
  */
 YaraRuleBuilder& YaraRuleBuilder::withStringMeta(const std::string& key, const std::string& value)
 {
-	if (key == std::string{} || value == std::string{})
-		throw RuleBuilderError("Error: String-Meta key and value must be non-empty.");
+	if (key.empty())
```
s3rvac: There is going to be a conflict when merging this PR and #59. I think it would be better to wait until #59 is in `master` and then only change `key == std::string{}` to `key.empty()` instead of replicating the work done in #59 (apart from a missing test for the bug fix).

src/builder/yara_rule_builder.cpp
```diff
@@ -365,7 +365,7 @@ YaraRuleBuilder& YaraRuleBuilder::withHexString(const std::string& id, const std
 {
 	createLastString();
 
-	if (id == std::string{} || hexString->getText() == std::string{})
+	if (id.empty() || hexString->getText().empty())
```
s3rvac: Consider adding a new member function to `hexString` called `empty()` or `is_empty()` to simplify the code and remove the [Law-of-Demeter violation](https://en.wikipedia.org/wiki/Law_of_Demeter) (more info in Czech is [here](https://cs-blog.petrzemek.net/2014-11-16-demeterin-zakon-aneb-minimalizujeme-strukturalni-zavislosti)).

Maybe I am taking things too far. I just wanted to provide an additional tip for consideration.

src/builder/yara_rule_builder.cpp
TadeasKucera: I have just resolved the conflict, it wasn't a tough one, but next time I can wait until the conflicting branch is merged. Thank you for your suggestion.

Issue #67, opened 2020-02-16
Handling of square brackets inside classes in regexes is broken


metthal:
When you have regexp like `/[[]*+]/` it handles it somehow weirdly and you end up with an error

```
yaramod.ParserError: Error at 8.28: Syntax error: Unexpected regexp +, expected one of (, ), /, regexp |, regexp ?, regexp ^, regexp $, regexp ., regexp ., regexp \w, regexp \W, regexp \s, regexp \S, regexp \d, regexp \D, regexp \b, regex
p \B, regexp class
```

If you remove the first `]` from the regexp (so `/[[*+]/``) it works. The square brackets inside classes are handled somehow weirdly. Even if you escape them `\]`. We don't have to build such a resistant parser that it will handle them when they are not escaped but I would expect that it doesn't fail when they are escaped.

PR #69, opened 2020-02-18
Implement parsing of nested regexp classes (issue #67)

TadeasKucera:
We are now able to parse nested regexp classes like `/[[]*+]/`, `/[\[\]*+]/` or even `/[[\]*+]/`.

### DIFF
--- a/include/yaramod/parser/parser_driver.h
+++ b/include/yaramod/parser/parser_driver.h
@@ -180,7 +180,7 @@ private:
 	std::string _indent; ///< Variable storing current indentation
 	std::string _comment; ///< For incremental construction of parsed comments
 	std::string _regexpClass; ///< Currently processed regular expression class.
-	size_t _regexpClassDepth = 0; ///< The rectangular brackets depth of nesting
+	std::size_t _regexpClassDepth = 0; ///< The rectangular brackets depth of nesting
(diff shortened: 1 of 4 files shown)


include/yaramod/parser/parser_driver.h
```diff
@@ -180,7 +180,9 @@ class ParserDriver
 	std::string _indent; ///< Variable storing current indentation
 	std::string _comment; ///< For incremental construction of parsed comments
 	std::string _regexpClass; ///< Currently processed regular expression class.
+	size_t _regexpClassDepth = 0; ///< The rectangular brackets depth of nesting
```
metthal: `std::size_t` instead of `size_t` to follow the conventions used in other parts of yaramod code.

src/parser/parser_driver.cpp
```diff
@@ -365,23 +365,37 @@ void ParserDriver::defineTokens()
...
+		--_regexpClassDepth;
+		if(_regexpClassDepth == 0)
+		{
+			enter_state("$regexp");
+			return std::make_pair(true, _regexpClass);
+		}
+		else
+		{
+			_regexpClass += "]";
+			return std::make_pair(false, "");
```
metthal: You can use `std::string{}` instead of `""` to denote empty string.

PR #77, opened 2020-03-10
Remove blank lines

TadeasKucera:
This feature removes double blank lines and use just one blank line instead when there is too many blank lines in the rule (Iff more than one third of the new-lines in the rule is followed by another new-line). The removing is only performed inside of rules.

### DIFF
--- a/src/types/token_stream.cpp
+++ b/src/types/token_stream.cpp
@@ -306,6 +306,7 @@ bool TokenStream::determineNewlineSectors()
 		else
 			wasLine = false;
 	}
+	// when more than half of the newlines is "doubled", then all "doubled" newlines are made simple newline.
 	return 3 * doubleLineCounter > lineCounter;
 }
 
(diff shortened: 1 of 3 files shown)


src/types/token_stream.cpp
```diff
@@ -291,6 +294,39 @@ void TokenStream::determineNewlineSectors()
...
+		{
+			++lineCounter;
+			if (wasLine)
+				++doubleLineCounter;
+			wasLine = true;
+		}
+		else
+			wasLine = false;
+	}
+	return 3 * doubleLineCounter > lineCounter;
```
TadeasKucera: This means that when more than half of the newlines is "doubled", then all "doubled" newlines are made simple new line.

metthal: This kind of comment would better fit directly in the source code :) Thanks.

metthal:
LGTM however I have a question. Does it only cover double or even triple, quadruple, .... blank lines?

TadeasKucera:
Thank you very much for noticing, this only covers double blank lines, because naturally the `std::list<T,Allocator>::erase` function returns iterator **after** the last removed element. Fixed in last commit.

PR #81, opened 2020-03-17
Align comments

TadeasKucera:
This PR implements autoformatting [improvement](https://github.com/avast/yaramod/wiki/Autoformatting#indenting-comments-inside-rule-sections) and also improves commented conjunctions / disjunctions produced in builder.

### DIFF
--- a/tests/cpp/parser_tests.cpp
+++ b/tests/cpp/parser_tests.cpp
@@ -4515,8 +4515,7 @@ import "cuckoo"
 rule public_rule
 {
 	condition:
-			(false and
-			true )
+			( /* comment */ false and true )
 }
 )");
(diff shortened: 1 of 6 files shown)


tests/cpp/parser_tests.cpp
```diff
@@ -4530,7 +4529,7 @@ rule public_rule
 {
 	condition:
 		(
-			false and
+			/* comment */ false and
 			true
 		)
```
TadeasKucera: This is one of the features of our autoformatter which has been designed not to be strict, but to be gentle. Of course the and conjunction forces a new line after and this new line forces new lines inside the parentheses sector. But we do not force any new lines after or before comments here. That is up to the user if they want to add some new lines or not.

PR #93, opened 2020-04-06
Fix parsing of "private global" rules

### LINKED ISSUE #92: Allow parsing of rules that are both private and global
First of all, let me say that I think this project is awesome and thanks for maintaining it!

Now the issue. Yara documentation is [quite clear](https://yara.readthedocs.io/en/v3.4.0/writingrules.html#private-rules) on the point that `private global` rules are allowed:

> You can apply both `private` and `global` modifiers to a rule, resulting a global rule that does not get reported by YARA but must be satisfied.

Indeed it works with yara:

```
[nix-shell:/tmp/test]$ cat test.yar
private global rule Kot
{
    strings:
        $a = "dummy1"

    condition:
        $a
}


rule Hmm
{
    condition:
        Kot
}


[nix-shell:/tmp/test]$ yara -rs test.yar .
Hmm ./test.yar
Hmm ./dummy
```

But this is not currently properly handled by yaramod:

```python
import yaramod

test = """
private global rule Kot
{
    strings:
        $a = "dummy1"

    condition:
        $a
}
"""

print(yaramod.Yaramod().parse_string(test).text_formatted)
```

```
[nix-shell:/tmp/test]$ python test.py
Traceback (most recent call last):
  File "test.py", line 14, in <module>
    print(yaramod.Yaramod().parse_string(test).text_formatted)
yaramod.ParserError: Error at 2.9-14: Syntax error: Unexpected global, expected one of rule
```

I couldn't find any issues similar to this one, so I've created a new one.

Please see my linked PR for proposed fix.

msm-code:
I've went ahead and tried to fix the issue I've found. I hope my C++ is adequate.

I tried to be careful not to introduce any regressions and memory leaks, but please do check.

I've also added a test to `parser_tests.cpp` but had problems building the test suite, so I just hope the test works. I tested my fixes by hand, using the python bindings instead:

```python
import yaramod

test = """
/* comment test */
private global rule Kot
{
    strings:
        $a = "dummy1"

    condition:
        $a
}
/* comment test */
global private rule Hmm
{
    strings:
        $a = "dummy1"

    condition:
        $a
}

/* this will not work:
global global rule Zzz
{
    strings:
        $a = "dummy1"

    condition:
        $a
}
*/

private rule private_rule {
    condition:
        true
}

global rule global_rule {
    condition:
        true
}
"""

print(yaramod.Yaramod().parse_string(test).text_formatted)

for rule in yaramod.Yaramod().parse_string(test).rules:
    print(rule.name, ("private" if rule.is_private else ""), ("global" if rule.is_global else ""))
```

This test doesn't work on current master branch (as expected), but works after this PR.

Fixes #92.

### DIFF
--- a/src/parser/parser_driver.cpp
+++ b/src/parser/parser_driver.cpp
@@ -430,7 +430,7 @@ void ParserDriver::defineGrammar()
 
 	_parser.rule("rule") // {}
 		.production(
-			"rule_mod", "RULE", [&](auto&&) -> Value {
+			"rule_mods", "RULE", [&](auto&&) -> Value {
 				_lastRuleLocation = currentFileContext()->getLocation();
 				_lastRuleTokenStream = currentFileContext()->getTokenStream();
 				return {};
@@ -441,7 +441,21 @@ void ParserDriver::defineGrammar()
 				return {};
 			},
 			"tags", "LCB", "metas", "strings", "condition", "RCB", [&](auto&& args) -> Value {
-				std::optional<TokenIt> mod = std::move(args[0].getOptionalTokenIt());
+				std::optional<TokenIt> mod_private = {};
+				std::optional<TokenIt> mod_global = {};
+				const std::vector<TokenIt> mods = std::move(args[0].getMultipleTokenIt());
+				for (const auto &token: mods) {
+					if (token->getType() == TokenType::GLOBAL) {
+						if (mod_global.has_value())
+							error_handle(token->getLocation(), "Duplicated global rule modifier");
+						mod_global = token;
+					}
+					else if (token->getType() == TokenType::PRIVATE) {
+						if (mod_private.has_value())
+							error_handle(token->getLocation(), "Duplicated private rule modifier");
+						mod_private = token;
+					}
+				}
 				TokenIt name = args[3].getTokenIt();
 				const std::vector<TokenIt> tags = std::move(args[5].getMultipleTokenIt());
 				args[6].getTokenIt()->setType(RULE_BEGIN);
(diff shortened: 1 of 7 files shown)


s3rvac:
Thank you for the very nicely written issue and PR! Since @metthal is on vacation, could I ask you @TadeasKucera to take a look at this PR and the associated issue #92?

TadeasKucera:
First of all thank you very much for this PR indeed. I have run the tests and all of them passed. I only have some minor remarks.

src/types/rule.cpp
```diff
@@ -335,7 +352,7 @@ void Rule::setLocation(const std::string& filePath, std::uint64_t lineNumber)
  */
 bool Rule::isGlobal() const
 {
-	return _mod.has_value() && (*_mod)->getType() == TokenType::GLOBAL;
+	return _mod_global.has_value();
```
TadeasKucera, line 355: This code is correct, but it expects, that when the rule is not global, the `_mod_global` is `std::nullopt`. The above change in the `Rule` constructor is therefore necessary.

src/parser/parser_driver.cpp
```diff
@@ -441,7 +441,21 @@ void ParserDriver::defineGrammar()
 				return {};
 			},
 			"tags", "LCB", "metas", "strings", "condition", "RCB", [&](auto&& args) -> Value {
-				std::optional<TokenIt> mod = std::move(args[0].getOptionalTokenIt());
+				std::optional<TokenIt> mod_private = {};
+				std::optional<TokenIt> mod_global = {};
+				const std::vector<TokenIt> mods = std::move(args[0].getMultipleTokenIt());
+				for (const auto &token: mods) {
```
TadeasKucera, line 447: Following the used code-style in yaramod, I would rather put the curly bracket `{` on a new line.

src/parser/parser_driver.cpp
```diff
@@ -441,7 +441,21 @@ void ParserDriver::defineGrammar()
...
+					if (token->getType() == TokenType::GLOBAL) {
```
TadeasKucera, line 448: Also here I would kindly ask you to put the bracket on a new line.

src/parser/parser_driver.cpp
```diff
@@ -441,7 +441,21 @@ void ParserDriver::defineGrammar()
...
+						if (mod_global.has_value())
+							error_handle(token->getLocation(), "Duplicated global rule modifier");
+						mod_global = token;
+					}
+					else if (token->getType() == TokenType::PRIVATE) {
```
TadeasKucera, line 453: If you don't mind, I would put the bracket on a new line.

src/types/rule.cpp
```diff
@@ -46,18 +46,33 @@ Rule::Rule(std::string&& name, Modifier mod, std::vector<Meta>&& metas, std::sha
 	}
 
 	if (mod == Modifier::Global)
-		_mod = _tokenStream->emplace_back(TokenType::GLOBAL, "global");
+	{
+		_mod_private = _tokenStream->emplace_back(TokenType::NONE, std::string{});
```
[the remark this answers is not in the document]
msm-code: :+1: Right, I just blindly followed the existing code instead of noticing that blank tokens are unnecessary now. Fixed.

src/parser/parser_driver.cpp
msm-code: Sure! I tried to be consistent with the style, but muscle memory won in this case.

msm-code:
No problem, glad I can help!

Issue #73, opened 2020-03-04
Improve links between individual constructs in internal representation


metthal:
Right now, the internal representation many times forgets link between individual construct in individual representation. One such example is this one

```yara
rule abc
{
        condition:
                true
}

rule def
{
        condition:
                abc
}
```

In this case, even though we check that `abc` must exist in symbol table while parsing condition of `def`, we forget this link and there is no way to get directly from `abc` in `def` condition to `Rule` instance of `abc` rule without looking up all rules in YARA file.

This ticket aims for improving the internal representation in such a way, that we'll keep these links and so it will be much easier to reach other parts of YARA file without doing whole file lookups.

**Use-case:**

I would like to rename rule and all links to it. So then if I do

```python
abc_rule.name = 'XYZ'
```

then I would expect output to be

```yara
rule XYZ
{
        condition:
                true
}

rule def
{
        condition:
                XYZ
}
```

and not 

```yara
rule XYZ
{
        condition:
                true
}

rule def
{
        condition:
                abc
}
```

These changes relate mostly to link between rules but I would also like to have something similar for strings.

**Technical details:**

It can be easier for rules since it can be solved just by some pointer to symbol table and renaming a rule would rename the record in symbol table (thus effectively renaming it everywhere else with zero cost) but it can be harder for strings since they are stored in trie-like structure for better lookup with string prefix. But that can possibly be solved by storing a pointer directly to the string itself which shouldn't change even if we move the string in trie to somewhere else.

TadeasKucera:
Solved in #96

PR #96, opened 2020-04-20
Improve links between individual constructs in internal representation (issue #73)

TadeasKucera:
In this PR we allow `yaramod::Literal` to reference another of it's kind so we can create links between tokens in a tokenstream. This brings a possibility to rename strings and rules easily. We had to move `yaramod::Token::Type` to a new file and make it a class `yaramod::TokenType` and similar with `yaramod::Expression::Type`.


include/yaramod/types/expression_type.h
```diff
@@ -0,0 +1,23 @@
...
+ * @brief Declaration of class ExpressionType.
+ * @copyright (c) 2017 Avast Software, licensed under the MIT license
+ */
+
+#pragma once
+
+namespace yaramod {
+
+///< Type of the expression.
+enum class ExpressionType
```
metthal, line 12: What was the motivation for taking out this type out of `Expression` itself?

include/yaramod/types/literal.h
```diff
@@ -96,6 +103,12 @@ class Literal
 			os << literal._formatted_value.value();
 		else if (literal.isBool())
 			os << (literal.getBool() ? "true" : "false");
+		else if (literal.isSymbol())
+			os << literal.getSymbol()->getName();
+		else if (literal.isLiteralReference())
+		{
```
metthal, line 109: Nitpick: there are no curly braces in other parts of this condition so I wouldn't expect them here either.

include/yaramod/types/expression_type.h
TadeasKucera: The `Literal` class now needs to access `Symbol::getName` method. That means it needs to include the _symbol.h_ file. In the `symbol.h` file there has been include of `expression.h` so we would end up with cyclic dependency (see picture). Luckily, the `Symbol` class only needs access to `ExpressionType` which has been extracted from `Expression::Type` to avoid this cyclic dependency.
yaramod_class_dependecy

include/yaramod/types/literal.h
TadeasKucera: I will fix it in my next PR, thank you for merging this.

Issue #75, opened 2020-03-06
Modifying rule with ModifyingVisitor should also modify it when obtaining getTextFormatted()


metthal:
This ticket is kinda related to #73 but it should also have its own because in some specific things it is a bit different.

Basically, we provide an option to modify the condition through `ModifyingVisitor` but these changes are not reflect when getting back the `text_formatted` out of `YaraFile`. This is caused by not directly modifying token streams when we are modifying the condition. For example following code:

```python
import json
import yaramod

class RegexpCaseInsesitiveAdder(yaramod.ModifyingVisitor):
    def add(self, yara_file: yaramod.YaraFile):
        for rule in yara_file.rules:
            new_condition = self.modify(rule.condition)
            print(new_condition.text)
            rule.condition = new_condition

    def visit_RegexpExpression(self, expr: yaramod.Expression):
        return yaramod.regexp('abc', 'i').get()

ymod = yaramod.Yaramod()
yfile = ymod.parse_file('/tmp/ruleset.yar')

regexp_icase_adder = RegexpCaseInsesitiveAdder()
regexp_icase_adder.add(yfile)

print(yfile.text_formatted)
```

This just replaces all regexes in conditions with `/abc/i` but when obtained through `text_formatted` it still return the original condition while it should return the modified one. It is expected that it returns the modified one.

TadeasKucera:
This issue was solved with the introduction of _ModifyingVisitor::cleanup_tokenstreams_ which allows us to write ModifyingVisitors that also alter the formatted text of affected rules.

PR #100, opened 2020-05-31
Fix acceptance of modifying visitor (#75)

TadeasKucera:
We introduce support for writing custom visitors modifying expressions. These new changes will make easier for authors of modifying visitors to create visitors which change the formatted text too:

- A new python `exchange_tokens` method of Expression class exchanges tokens of the expression with tokens stored in its parameter.
- Expression methods getFirstTokenIt, getLastTokenIt are introduced.
- Many visitor examples are created in new _tests/cpp/visitor_tests.cpp_

These changes will help us write functional modifying visitors. I believe this can be a good solution to #75, and the visitor mentioned in the issue can be rewritten as follows:

```
class RegexpCaseInsesitiveAdder(yaramod.ModifyingVisitor):
    def add(self, yara_file: yaramod.YaraFile):
        for rule in yara_file.rules:
            self.modify(rule.condition)

    def visit_RegexpExpression(self, expr: yaramod.Expression):
         output = yaramod.regexp('abc', 'i').get()
         expr.exchange_tokens(output)
         return output
```
Do you agree with this solution? Many visitor examples can be seen in the `visitor_tests.cpp` file.

Afterward we improved the provided interface to offer the following functionality:

- There is a `TokenStreamContext` class which is just context storage created from `Expression` instance when needed (typically at the beginning of an `accept` method which will modify the expression).
- More importantly, the `cleanUpTokenStreams(TokenStreamContext context, Expression* e)` extracts TokenStream `ts` from `context` and then deletes all previous tokens of `e` from `ts` and moves the new tokens from the TokenStream belonging to expression `e` into `ts`. This is, of course, done only when the `e` has a new TokenStream - other than `ts` (which typically means it was created by a builder).

For example, the following visitor will replace all `==` with `!=` and also change the order of the `EqExpression` parameters order. Therefore `!str1 == !str2` becomes `!str2 != !str1`:

```
class EqModifyer(yaramod.ModifyingVisitor):
    def add(self, yara_file: yaramod.YaraFile):
        for rule in yara_file.rules:
            rule.condition = self.modify(rule.condition)

    def visit_EqExpression(self, expr: yaramod.Expression):
        context = yaramod.TokenStreamContext(expr)
        expr.left_operand.accept(self)
        expr.right_operand.accept(self)
        output = (yaramod.YaraExpressionBuilder(expr.right_operand) != yaramod.YaraExpressionBuilder(expr.left_operand)).get()

        self.cleanUpTokenStreams(context, output)
        return output
```


src/python/yaramod_python.cpp
```diff
@@ -297,6 +297,9 @@ void addExpressionClasses(py::module& module)
 	py::class_<Expression, std::shared_ptr<Expression>>(module, "Expression")
 		.def("accept", &Expression::accept)
 		.def("get_text", &Expression::getText, py::arg("indent") = std::string{})
+		.def("exchange_tokens", [](Expression& self, Expression* other) {
```
metthal, line 300: Won't `&Expression::exchangeTokens` be sufficient here instead of lambda?

TadeasKucera: The `Expression::exchangeTokens` method has four overloads but I can put there this: 
`py::overload_cast<Expression*>(&Expression::exchangeTokens)`

TadeasKucera: As for the second thing, I just pushed some [commits](https://github.com/avast/yaramod/pull/100/commits/79fde8d440be9192a6e0ca755a7dd6ff7bd60a46) with new `OrExpressionSwitcher` modifying visitor which may be a solution to this. This visitor has been written in a manner in which the dealing with tokens can be made 1. before and 2. at-the-end of the `visit` action. I will test this little further.

include/yaramod/types/expression.h
```diff
@@ -71,19 +71,23 @@ class Expression
 			default: return "Error - unknown type";
 		}
 	}
-	TokenStream* getTokenStream()
+	TokenStream* getTokenStream() const
 	{
 		return _tokenStream.get();
 	}
-	void setTokenStream(const std::shared_ptr<TokenStream>& ts)
+	const std::shared_ptr<TokenStream> getTokenStreamSharedPtr() const
```
metthal, line 78: You are returning a copy here. You most probably wanted to return the reference.

include/yaramod/types/expressions.h
```diff
@@ -1199,7 +1242,7 @@ class IdExpression : public Expression
 	 * Constructors
 	 */
 	IdExpression(const std::shared_ptr<Symbol>& symbol)
-		: _symbol(symbol)
+		: _symbol(symbol) //TODO
```
metthal, line 1245: What does this TODO mean?

include/yaramod/types/expressions.h
```diff
@@ -1325,6 +1380,10 @@ class ArrayAccessExpression : public IdExpression
 	const Expression::Ptr& getArray() const { return _array; }
 	const Expression::Ptr& getAccessor() const { return _accessor; }
 
+	//TODO: check if is correct:
```
metthal, line 1383: Should this still be here?

include/yaramod/types/expressions.h
```diff
@@ -1387,6 +1446,10 @@ class FunctionCallExpression : public IdExpression
 	const Expression::Ptr& getFunction() const { return _func; }
 	const std::vector<Expression::Ptr>& getArguments() const { return _args; }
 
+	//TODO: check if is correct:
```
metthal, line 1449: Same here

include/yaramod/types/expressions.h
```diff
@@ -1387,6 +1446,10 @@ class FunctionCallExpression : public IdExpression
...
+	virtual TokenIt getFirstTokenIt() const override { return _func->getFirstTokenIt();  /*return _symbolToken.value();*/ }
```
metthal, line 1450: What does the commented out code mean?

include/yaramod/utils/modifying_visitor.h
```diff
@@ -10,6 +10,34 @@
 #include "yaramod/utils/visitor.h"
 
 namespace yaramod {
+/**
+ * This class captures current TokenStream and first and last TokenIi of the expression which it is given
+ * in it's constructor. This data can be very useful at the end of the visit method of ModifyingVisitors
+ * by the cleanUpTokenStreams function.
+ */
+class TokenStreamContext {
```
metthal, line 18: Please follow code style used in other files

include/yaramod/utils/modifying_visitor.h
```diff
@@ -10,6 +10,34 @@
...
+	{
+	}
+	TokenStreamContext(const std::shared_ptr<TokenStream>& oldTS, TokenIt oldBeforeFirst, TokenIt oldAfterLast)
+		: _oldTS(oldTS)
+		, _oldBeforeFirst(oldBeforeFirst)
+		, _oldAfterLast(oldAfterLast)
+	{
+	}
+
+	std::shared_ptr<TokenStream>& oldTS() { return _oldTS; }
```
metthal, line 33: I would prefer `oldTokenStream` instead of `oldTS` since to me it seems like unnecessary abbreviation.

src/types/token_stream.cpp
```diff
@@ -19,87 +19,198 @@ constexpr unsigned tabulator_length = 8;
...
 }
 
 void TokenStream::move_append(TokenStream* donor)
 {
 	_tokens.splice(_tokens.end(), donor->_tokens);
+	_formatted = false;
 }
 
-void TokenStream::move_append(TokenStream* donor, TokenIt before)
+void TokenStream::move_append(TokenIt before, TokenStream* donor)
```
metthal, line 108: What I noticed recently is that we are mixing here camel case which is usually established in yaramod however there are these methods with snake casing. Not something crucial but it would be nice to stick to single naming style.

include/yaramod/types/expression.h
TadeasKucera: Thank you, I missed the & here.

include/yaramod/types/expressions.h
TadeasKucera: Thank you for noticing. This TODO means to check that any IdExpression (or derived class) instance will have its `_symbolToken` valid TokenIt after parsing. I have add test to test this and it seems alright.

include/yaramod/types/expressions.h
TadeasKucera: No, this can be removed. I am sorry I ignored all those `TODO`. I removed it and it should be fine - thank you.

include/yaramod/types/expressions.h
TadeasKucera: It is old code which is now wrong. I removed it.

include/yaramod/utils/modifying_visitor.h
TadeasKucera: Agreed, I fixed it :-)

include/yaramod/utils/modifying_visitor.h
TadeasKucera: I added a new line in between class name and the curly bracket and also one more new line before `private` keyword. I hope it is alright now.

src/types/token_stream.cpp
TadeasKucera: I am not sure whether we also want to rename `TokenStream::emplace_back` to `TokenStream::emplaceBack`. Also `pushBack` looks a bit strange when underlying List has these methods named in snake case. I would probably have big troubles to get used to it. Do you mind if we keep it `emplace_back` and `push_back`?

But I renamed `move_append` to `moveAppend` because standard c++ containers have no such methods and therefore it should not cause us troubles to get used to camel case in this case.

TadeasKucera: Also `swap_tokens` will be renamed to `swapTokens`.

include/yaramod/types/expressions.h
TadeasKucera: There should be no more TODOs in yaramod (in our code) now.

TadeasKucera:
Tomorrow I will add the documentation as well.

docs/rtd/modifying_rulesets.rst
```diff
@@ -0,0 +1,121 @@
...
+  def visit_RegexpExpression(self, expr: yaramod.Expression):
+    output = yaramod.regexp('abc', 'i').get()
+    expr.exchange_tokens(output)
+    return output
+
+This ``visit`` methods requires calling of a ``exchange_tokens`` method which deletes all Tokens that the original expression refered to. Then it extracts all tokens from the supplied new Expression and moves them to place where the original expression had its tokens stored.
+
+In the third example we will show how to deal with a situation where we need to modify existing expression while keeping part of it's subexpressions. The following approach will let us use Yaramod Expression Builders to create new Expressions from existing expressions that are already used in our rule.
+
+Let's now assume that we need to modify each `EqExpression` in the expression syntax tree. We can do it by writing our own derived class of `ModifyingVisitor`. The new class will override the `visit(EqExpression* expr)` method in the following manner:
```
metthal, line 105: These are probably not correctly escaped as double ` are needed.

metthal:
Could you also please rebase the branch as I've added checking of documentation build into Travis CI pipeline to ensure that the modified documentation is buildable.

TadeasKucera:
I have added the c++ snippets to the documentation and also made tests for them. Thank you for adding the documentation build to Travis pipeline, which is very handy.

Issue #99, opened 2020-05-25
Allow access to Token, TokenStream and Location also in python bindings


TadeasKucera:
The `Token`, `TokenStream` and `Location` classes are now not accessible through python bindings, which should change. We will also need to add an python binding for the method `YaraFile::getTokenStream`

PR #101, opened 2020-06-18
Access ts after syntax error (#99)

TadeasKucera:
After throwing ParserError, we can access the parsed YaraFile instance directly through `ParserDriver` instance or newly also from `Yaramod` class instance. The YaraFile class owns a valid TokenStream despite the fact that the ParserError has been thrown. Access to this TokenStream is now tested both in `parser_tests.cpp` and also in `representation_tests.py` through new python bindings:
Invalid rule
```
	rule dummy_rule {
		condition:
			true ) and false
	}
```
will fail to be parsed, but the TokenStream still remains accesible and contains some valuable data:
```
        ymod = yaramod.Yaramod()
        try:
            ymod.parse_string(input_text)
        except:
            ts = ymod.yara_file.tokenstream
            self.assertFalse(ts.empty)
            self.assertEqual(ts.front.pure_text, '\n')
            self.assertEqual(ts.back.pure_text, ')')
```
This PR also partially solves #99. We implement access to `Token` and `TokenStream` classes in python.


src/python/yaramod_python.cpp
```diff
@@ -117,6 +117,7 @@ void addBasicClasses(py::module& module)
 		.def_property_readonly("rules", &YaraFile::getRules)
 		.def_property_readonly("imports", &YaraFile::getImports)
 		.def_property_readonly("text_formatted", [](const YaraFile& self) { return self.getTextFormatted(); })
+		.def_property_readonly("tokenstream", [](const YaraFile& self) { return *(self.getTokenStream());} )
```
metthal, line 120: Does that mean that accessing `tokenstream` will make copy of the whole tokenstream or will it return by reference?

TadeasKucera: This makes a copy of the `TokenStream` which is not the desired behaviour. Thank you for pointing it out!

TadeasKucera: I fixed it in the last commit [a113ee3](https://github.com/avast/yaramod/pull/101/commits/a113ee3b69c14490b7d815f8902f8ff552826aff). I tested that it returns by reference now.

PR #113, opened 2020-07-20
Added support for classes and windows

stepanek-m:
Added parsing of functions `cuckoo.process.class_created()`, `cuckoo.process.class_searched()`, `cuckoo.process.window_created()` and `cuckoo.process.window_searched()`.

### DIFF
--- a/tests/cpp/parser_tests.cpp
+++ b/tests/cpp/parser_tests.cpp
@@ -6480,6 +6480,94 @@ rule cuckoo_desktop
 }
 
 TEST_F(ParserTests,
+CuckooClassCreatedModuleFunction) {
+	prepareInput(
+R"(
+import "cuckoo"
+
+rule cuckoo_class_created
+{
+	condition:
+		cuckoo.process.class_created(/SomeEvilClass/)
+}
+)");
+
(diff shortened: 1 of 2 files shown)


tests/cpp/parser_tests.cpp
```diff
@@ -6479,6 +6479,94 @@ rule cuckoo_desktop
...
+
+	EXPECT_TRUE(driver.parse(input));
+	ASSERT_EQ(1u, driver.getParsedFile().getRules().size());
+
+	const auto& rule = driver.getParsedFile().getRules()[0];
+	EXPECT_EQ(R"(cuckoo.process.window_searched(/SomeEvilWindow/))", rule->getCondition()->getText());
+
+	EXPECT_EQ(input_text, driver.getParsedFile().getTextFormatted());
+}
+
```
TadeasKucera, line 6569: In all these tests, we use the `driver` instance of `ParserDriver` that has been constructed with default parameter `ImportFeatures::AllCurrent`.
Maybe we would also like to test, that the behavior would be the same if we used an instance of `ParserDriver` class constructed with `ImportFeatures::Avast`, and, that when we would use `ImportFeatures::VirusTotal`, the parsing would fail.

PR #120, opened 2020-08-04
Token locations python bindings

MatejKastak:
Add a way to access locations of various objects in yara files from python.

## Interface
- `location` is now represented as a [structure](https://github.com/avast/yaramod/pull/120/files#diff-adf2b1173ddb83b9ea215bc334a3ffbaR22) instead of `std::pair`
- keeping the old `Rule::Location` interface, `line_number` and `file_path`
- exposing `TokenType` for each `Token`
- extending the `String` interface to provide `location` attribute just like the rules
- exposing some tokens linked to symbols
    - `Rule` - `token_first`, `token_last`
    - `String` - `token_first`, `token_last`, `token_id`, `token_assign`

## TODO

- [ ] 5e6dc9c6ddd01638cc75b92239fdad43f966e890 contains changes that convert `TokenType` to cpp11 style `enum class` enums, I reverted them because I think they are out of the scope of this PR, if you want I can open a separate PR for just this. What do you think about this change?
- [ ] I provided a [baseline for the tests](https://github.com/avast/yaramod/pull/120/files#diff-cb9bfd7a40553f489a0e1dae2d724f7cR1611), but currently some of the token locations are not correct. I have marked them with `FIXME:` [comments](https://github.com/avast/yaramod/pull/120/files#diff-cb9bfd7a40553f489a0e1dae2d724f7cR1662). Keep in mind that the values in the tests are my expectations, I you think they are not right we can discuss the changes.


TadeasKucera:
First of all thank you for the PR and for recognizing the problems with location of some tokens. The code is alright and I approve it. As for the first checkbox: Please do open a separate pull request for it.

PR #124, opened 2020-08-11
Fix token location computing

TadeasKucera:
Fixing the location of tokens of strings.

- [x] Plain strings
- [x] Hex Strings
- [ ] Regexp strings -- this shall be solved in #125.

### DIFF
--- a/include/yaramod/parser/location.h
+++ b/include/yaramod/parser/location.h
@@ -21,13 +21,19 @@ public:
 	 */
 	struct Position {
 
+		Position() : line(1), column(0) {}
 		Position(std::size_t line, std::size_t column) : line(line), column(column) {}
 
-		std::size_t getLine() { return line; }
-		std::size_t getColumn() { return column; }
+		std::size_t getLine() const { return line; }
+		std::size_t getColumn() const { return column; }
 
 		std::size_t line;
 		std::size_t column;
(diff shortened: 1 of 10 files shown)


tests/python/test_parser.py
```diff
@@ -1646,21 +1646,21 @@ def test_string_locations(self):
...
         self.assertEqual(string1.token_last.location.end.line, 3)
         self.assertEqual(string1.token_last.location.end.column, 27)
 
         string2 = rule.strings[1]
         self.assertEqual(string2.identifier, '$2')
         self.assertEqual(string2.location.begin.line, 4)
         self.assertEqual(string2.location.begin.column, 10)
-        # self.assertEqual(string2.location.end.line, 4)  # FIXME: 4 != 8
-        # self.assertEqual(string2.location.end.column, 40)  # FIXME: 2 != 40
+        self.assertEqual(string2.location.end.line, 4)  # FIXME: 4 != 8
```
metthal, line 1662: Please remove `FIXME` comments

include/yaramod/parser/location.h
```diff
@@ -21,13 +22,19 @@ class Location
 	 */
 	struct Position {
 
+		Position() : line(1), column(0) {}
```
MatejKastak, line 25: I think it would be beneficial to add comment about how we count lines and columns.

- lines from 1
- columns from 0

PR #129, opened 2020-08-18
Fix include directive not to extract tokens

TadeasKucera:
This PR makes sure that `YaraFile::textFormatted` does not include tokens of included files. We also add a test with first real files.


src/types/yara_file.cpp
```diff
@@ -211,19 +211,20 @@ bool YaraFile::addImport(TokenIt import, ModulesPool& modules)
  *
  * @param rule Rule to add.
```
MatejKastak, line 212:
```suggestion
 * @param rule Rule to add.
 * @param extractTokens set if the tokens should be extracted into the _tokenStream.
```

tests/python/test_visitor.py
```diff
@@ -687,3 +686,86 @@ def visit_FunctionCallExpression(self, expr):
...
+	strings:
+		$str1 = "a"
+	condition:
+		INTERESTING and
+		(
+			$str1
+		)
+}
+'''
+        self.assertEqual(expected, yara_file.text_formatted)
```
MatejKastak, line 771:
```suggestion
        self.assertEqual(expected, yara_file.text_formatted)
        
```

MatejKastak (approved):
I was able to test your patch and the problem is not present anymore :+1:

- regarding the code, I left some nitpicks that are up to your consideration

PR #131, opened 2020-08-21
Add test of RuleInserter

TadeasKucera:
I have added one more test of ModifyingVisitor. This one attempts to create a new rule `ANOTHER_RULE` and add it into an existing file. It also modifies each present rule to refer to the newly created `ANOTHER_RULE`.


MatejKastak (approved):
I see no problems with this test, and it also passed.

- from the code I can see that `ModifyingVisitor` is really powerful tool :+1:

PR #148, opened 2020-11-01
Variables and bool arrays

tomaskender:
Add support for new yara constructs.

variable:
x = 5+5
\------------------
any of <iterator>

* parser and builder support for new features
* adds support for <iterator>, which is a set with square brackets for any kind of expressions. more potential use-cases for 'in' operator in the future
* adds support for referencing symbols in local symbol table under different names
* tests


include/yaramod/types/expressions.h
```diff
@@ -1169,6 +1169,70 @@ class OfExpression : public ForExpression
...
+ * Class representing an iterator, which is an array of expressions typically
+ * used with an of operator
+ *
+ * For example:
+ * @code
+ * all of [ true, false, true ]
+ *        ^^^^^^^^^^^^^^^^^^^^^
+ * @endcode
+ */
+class IteratorExpression : public Expression
```
metthal, line 1182: Is _iterator_ the correct name to choose for this abstraction? _Iterator_ is usually something which iterates over _iterable_. The difference here is that iterator carries notion of what is current element and which one is the next one. This sequence doesn't carry anything like this so is it really `Iterator` or rather `Iterable`?

metthal (changes requested):
1 more questions & 2 more task:

Question: Do I understand it correctly that there's no way to use `YaraExpressionBuilder` together with newly introduced variables? From what I see, you can only initialize variables to some static values (strings, integers, floats, ...) but apart from that, you can't for example build `1 + 2` using builder interface. It's something that should be improved but we don't have a use-case for it right now so we'll be fine without it.

Task 1: Variables and iterables are going to be Avast-only feature for some time (until upstream decides what to do with our changes) so I would expect that these will be enabled only when Avast-mode is enabled. When you are creating new `Yaramod` instance, you are able to provide it something called `ImportFeatures`. I would rename this to `Features` and also include grammar modifications in there.

Task 2: Please also add documentation for these new features in `docs/rtd`. Both building and parsing rulesets. Don't forget to also change references to `ImportFeatures` and mentioning them in description of these new features.

include/yaramod/types/symbols.h
```diff
@@ -185,4 +185,18 @@ class StructureSymbol : public Symbol
...
+
+	const std::shared_ptr<Symbol>& getSymbol() const {
+		return _symbol;
+	}
+
+private:
+	std::shared_ptr<Symbol> _symbol;
+};
+
+}
```
TadeasKucera, line 202: Please add a new line here.

include/yaramod/types/symbols.h
```diff
@@ -185,4 +185,18 @@ class StructureSymbol : public Symbol
...
 };
 
-}
+
+class ReferenceSymbol : public Symbol
+{
+public:
+	ReferenceSymbol(const std::string& name, const std::shared_ptr<Symbol>& symbol) : Symbol(Symbol::Type::Reference, name, ExpressionType::Object), _symbol(symbol) {}
+
+	const std::shared_ptr<Symbol>& getSymbol() const {
```
TadeasKucera, line 194: To follow used code style I would suggest to move the { on a new separate line.

TadeasKucera (changes requested):
I have come through the code and added some suggestions, mostly regarding the code style. Also I would suggest to add a few particular tests (see my comments). Despite that the code seems tidy. Thank you!

In addition to the tests described in the comments I would consider adding another c++ test to `tests/cpp/parser_tests.cpp` which would check, that a rule with variables that are also addressed from its condition has a proper formatting. This [test](https://github.com/avast/yaramod/blob/3f874f9f20c5ba9d3e9145a828ccfb89cadf24fe/tests/cpp/parser_tests.cpp#L6332) might be an inspiration.

src/parser/parser_driver.cpp
```diff
@@ -118,6 +118,7 @@ void ParserDriver::defineTokens()
 	_parser.token("private").symbol("PRIVATE").description("private").action([&](std::string_view str) -> Value { return emplace_back(TokenType::PRIVATE, std::string{str}); });
 	_parser.token("rule").symbol("RULE").description("rule").action([&](std::string_view str) -> Value { return emplace_back(TokenType::RULE, std::string{str}); });
 	_parser.token("meta").symbol("META").description("meta").action([&](std::string_view str) -> Value { return emplace_back(TokenType::META, std::string{str}); });
+	_parser.token("variables").symbol("VARIABLES").description("variables").action([&](std::string_view str) -> Value { return emplace_back(TokenType::VARIABLES, std::string{str}); });
```
metthal, line 121: Would it be possible to have this token inside of AvastOnly parsing enabled?

src/parser/parser_driver.cpp
```diff
@@ -447,7 +448,7 @@ void ParserDriver::defineGrammar()
 				args[3].getTokenIt()->setValue(std::make_shared<ValueSymbol>(name_text, Expression::Type::Bool));
 				return {};
 			},
-			"tags", "LCB", "metas", "strings", "condition", "RCB", [&](auto&& args) -> Value {
+			"tags", "LCB", "metas", "strings", "variables", "condition", "RCB", [&](auto&& args) -> Value {
```
metthal, line 451: Would it be possible to have 2 possible productions based on whether AvastOnly parsing is enabled or not? Something like this

```c++
auto& rule = _parser.rule("rule");
if (avast_parser_enabled) {
    rule.production(production_with_variables);
} else {
    rule.production(production_without_variables);
}
```

I prefer keeping the grammar as it is if we don't use any additional features.

metthal (changes requested):

src/parser/parser_driver.cpp
```diff
@@ -433,51 +435,96 @@ void ParserDriver::defineGrammar()
...
-				std::shared_ptr<Rule::StringsTrie> strings = std::move(args[8].getStringsTrie());
-				Expression::Ptr condition = std::move(args[9].getExpression());
-				args[10].getTokenIt()->setType(TokenType::RULE_END);
-
-				addRule(Rule(_lastRuleTokenStream, name, std::move(mod_private), std::move(mod_global),
-					std::move(metas), std::move(strings), std::move(condition), std::move(tags)));
-				return {};
-			});
+  auto const common_last_rule = [&](auto&&) -> Value {
+  	_lastRuleLocation = currentFileContext()->getLocation();
```
metthal, line 439: Seems like an indentation inconsistency.

src/parser/parser_driver.cpp
```diff
@@ -433,51 +435,96 @@ void ParserDriver::defineGrammar()
...
+		args[3].getTokenIt()->setType(TokenType::RULE_NAME);
+		args[3].getTokenIt()->setValue(std::make_shared<ValueSymbol>(name_text, Expression::Type::Bool));
+		return {};
+	};
+
+  if (_features & Features::AvastOnly)
+    _parser.rule("rule") // {}
+		  .production(
+			  "rule_mods", "RULE", common_last_rule, "ID", common_rule_init, "tags", "LCB", "metas", "strings", "variables", "condition" , "RCB", [&](auto&& args) -> Value {
+	        std::optional<TokenIt> mod_private = {};
```
metthal, line 457: Check if you are using tabs/spaces consistently. Some lines are aligned differently than others.

src/parser/parser_driver.cpp
```diff
@@ -1156,6 +1237,18 @@ void ParserDriver::defineGrammar()
 			output->setTokenStream(currentTokenStream());
 			return output;
 		})
+		.production("for_expression", "OF", "expression_iterable", [this](auto&& args) -> Value {
```
metthal, line 1240: Shouldn't also this in a condition that checks for Avast features?

src/parser/parser_driver.cpp
```diff
@@ -1766,6 +1869,34 @@ void ParserDriver::defineGrammar()
 			return output;
 		})
 		;
+		
+	_parser.rule("expression_iterable") // Expression::Ptr
```
metthal, line 1873: Shouldn't also this be in a condition that checks for Avast features?

src/parser/parser_driver.cpp
```diff
@@ -433,51 +435,96 @@ void ParserDriver::defineGrammar()
...
+      })
+      ;
+  else
+	  _parser.rule("rule") // {}
+		  .production(
+			  "rule_mods", "RULE", common_last_rule, "ID", common_rule_init, "tags", "LCB", "metas", "strings", "condition" , "RCB", [&](auto&& args) -> Value {
+		      std::optional<TokenIt> mod_private = {};
+  		    std::optional<TokenIt> mod_global = {};
+	  	    const std::vector<TokenIt> mods = std::move(args[0].getMultipleTokenIt());
+		      for (const auto &token: mods)
```
metthal, line 500: This is kinda better in a sense that at least it's in condition for Avast features but can we at least simplify this solution so that we don't repeat the same code? If you are not able to do that via `std::bind` and lambdas then at least please introduce some free functions that would do these bits atomically and we can reuse them? Imagine that there's a bug in this for loop for some reason. You'd have to change it on multiple places and that is heavily error prone. Thank you.

metthal (changes requested):

PR #152, opened 2021-01-15
Fix Trie::getValueWithPrefix to return values in the correct order

TadeasKucera:
Fixed the order of results from `Trie::getValueWithPrefix`.

### DIFF
--- a/include/yaramod/utils/trie.h
+++ b/include/yaramod/utils/trie.h
@@ -368,7 +368,7 @@ public:
 				nodes.push_back(subnode);
 		}
 
-		return result;
+		return {result.rbegin(), result.rend()};
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


include/yaramod/utils/trie.h
```diff
@@ -368,7 +368,7 @@ class Trie
 				nodes.push_back(subnode);
 		}
 
-		return result;
+		return {result.rbegin(), result.rend()};
```
metthal, line 371: What about using `std::reverse` and reversing the `result` in place? I am not sure about RVO picking on this and eliminating unnecessary allocations.

metthal (changes requested):

include/yaramod/utils/trie.h
TadeasKucera: That is certainly a better idea, thank you and please have a look once again.

metthal: LGTM now. And thanks for noticing and fixing it.

metthal (approved):

Issue #45, opened 2019-12-11
Allow parsing of incomplete rules


metthal:
It would be nice to parse rules also in some kind of `IncompleteRules` mode which would ignore that some symbols might be missing from symbol tables. This would allow us to for example autoformat rules even without any includes which refer to other rules that are not part of the autoformatted rule.

For example let's say I have

```
rule abc {
    condition:
        xyz
}
```

`xyz` is not defined here but it might come from some include later therefore we'll just parse it silently and ignore it.

metthal:
This should also allow us to obtain token stream from the parsed YARA file whenever there's a syntax error. For example let's say I have

```
rule abc {
```

This is not valid YARA but I can obtain token stream and by looking at the stream in reverse, I can infer that the next token should be `meta` and advise it to the user.

metthal:
Another possible feature of this should be to ignore includes of files it can't find.

PR #159, opened 2021-05-18
Parsing of incomplete rules

TadeasKucera:
This PR solves issue [#45](https://github.com/avast/yaramod/issues/45) and allows parsing of:

-  incomplete rules (we can get TokenStream after parsing)
-  rules containing unknown identifiers (we can get text, text_formatted and condition)
-  YARA files importing non-existent files (we can get text, text_formatted and condition)

### DIFF
--- a/include/yaramod/types/symbol.h
+++ b/include/yaramod/types/symbol.h
@@ -34,7 +34,8 @@ public:
 		Dictionary,
 		Function,
 		Structure,
-		Reference
+		Reference,
+		Undefined
 	};
 
 	/// @name Destructor
@@ -58,6 +59,7 @@ public:
 			case Type::Function : return TokenType::FUNCTION_SYMBOL;
 			case Type::Structure : return TokenType::STRUCTURE_SYMBOL;
 			case Type::Reference : return TokenType::REFERENCE_SYMBOL;
+			case Type::Undefined : return TokenType::UNDEFINED;
 			default: return TokenType::INVALID;
 		}
 	}
@@ -77,6 +79,7 @@ public:
 	bool isFunction() const { return _type == Symbol::Type::Function; }
 	bool isStructure() const { return _type == Symbol::Type::Structure; }
 	bool isReference() const { return _type == Symbol::Type::Reference; }
+	bool isUndefined() const { return _type == Symbol::Type::Undefined; }
 	/// @}
 
 protected:
--- a/include/yaramod/types/symbol.h
+++ b/include/yaramod/types/symbol.h
@@ -38,6 +38,14 @@ public:
 		Undefined
 	};
 
+	/// @name Constructors
+	/// @{
+	Symbol(Symbol::Type type, const std::string& name, ExpressionType dataType, const std::string& documentation = "")
+		: _type(type), _name(name), _documentation(documentation), _dataType(dataType)
+	{
+	}
+	/// @}
+
 	/// @name Destructor
 	/// @{
 	virtual ~Symbol() = default;
@@ -83,13 +91,6 @@ public:
 	/// @}
 
 protected:
-	/// @name Constructors
-	/// @{
-	Symbol(Symbol::Type type, const std::string& name, ExpressionType dataType, const std::string& documentation = "")
-		: _type(type), _name(name), _documentation(documentation), _dataType(dataType)
-	{
-	}
-	/// @}
 
 	Symbol::Type _type; ///< Type of the symbol
 	std::string _name; ///< Name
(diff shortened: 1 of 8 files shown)


include/yaramod/types/symbol.h
```diff
@@ -34,9 +34,18 @@ class Symbol
...
+		Undefined
 	};
 
+	/// @name Constructors
+	/// @{
+	Symbol(Symbol::Type type, const std::string& name, ExpressionType dataType, const std::string& documentation = "")
+		: _type(type), _name(name), _documentation(documentation), _dataType(dataType)
+	{
+	}
+	/// @}
```
TadeasKucera, line 47: I needed this constructor public, so I could construct a new `Symbol` instance in `parser_driver.cpp` on line 1635.

src/parser/parser_driver.cpp
```diff
@@ -12,6 +12,8 @@
 #include "yaramod/types/token_type.h"
 #include "yaramod/utils/filesystem_operations.h"
 
+#include <iostream>
```
TadeasKucera, line 15: I started printing through `std::cerr` instead of throwing `ParserError` when parsing fails in `IncompleteMode`.This only happens if a YARA rule is not finished (missing `}`, etc.). It prints the error as Warning but we can still access the `TokenStream` as we can see in `ParseIncompleteRuleNotFinished` cpp parser test.

src/parser/parser_driver.cpp
```diff
@@ -1067,7 +1075,16 @@ void ParserDriver::defineGrammar()
...
+				{
+					if (!incompleteMode() || !parentSymbol->isUndefined())
+						error_handle((++args[3].getTokenIt())->getLocation(), "Identifier '" + parentSymbol->getName() + "' is not an array");
+					else
+					{
+						auto parentExpr = std::static_pointer_cast<IdExpression>(iterable);
+						const auto& parentTokenText = parentExpr->getSymbolToken()->getText();
+						parentExpr->setSymbol(std::make_shared<ArraySymbol>(parentTokenText, ExpressionType::Undefined));
+					}
+				}
```
TadeasKucera, line 1087: Why so complicated? I needed the tormatted text to still look nice, which means I had to replace some `Symbol` base class instances with instances of particular derived classes. Also it was necessary to determine correct types of some Token instances (not everywhere).

TadeasKucera: Good thing is, that the `IdExpression::setSymbol` method also updates the pointer stored in the `Token` associated with this `Symbol` instance.

src/parser/parser_driver.cpp
```diff
@@ -1936,7 +2023,18 @@ bool ParserDriver::parse(std::istream& stream, ParserMode parserMode)
...
+	try {
+		auto output = parseImpl();
+		return output;
+	} catch (const ParserError& e) {
+		if (!incompleteMode())
+			throw e;
+		else
+		{
+			std::cerr << "Warning: '" << e.getErrorMessage() << "'\nThis Warning would be an Error if not parsing in Incomplete mode." << std::endl;
+			return true;
```
TadeasKucera, line 2035: I would appreciate any reviewer's opinien on if this printout is actually wanted. Thank you in advance.

tests/cpp/parser_tests.cpp
```diff
@@ -5204,6 +5204,354 @@ rule rule_3
...
+}
+)"
+);
+	EXPECT_TRUE(driver.parse(input, ParserMode::Incomplete));
+	ASSERT_EQ(1u, driver.getParsedFile().getRules().size());
+	ASSERT_EQ(input_text, driver.getParsedFile().getTextFormatted());
+}
+
+TEST_F(ParserTests,
+ParseIncompleteRuleNotFinished) {
```
TadeasKucera, line 5541: This is how incomplete rule can be parsed. The `TokenStream` instance is still accessible.

src/parser/parser_driver.cpp
metthal: Printing from libraries is generally bad practice. It's much better if library returns the errors/warnings to the user and they can decide what to do with them. Imagine a situation that you don't want to print to `stderr` or simply `stderr` is missing because the application is daemon. How would you know about such error/warning?

src/parser/parser_driver.cpp
metthal: As I said earlier, let the user handle the prints themselves. But I'd argue here that it's pointless to show something like this to an user. Incomplete mode is opt-in so user must know what they are doing.

metthal (changes requested):

TadeasKucera:
Thank you for your comments. I removed the Warning when ParsingError for the reason you stated.

src/parser/parser_driver.cpp
TadeasKucera: I decided to stop printing these Warnings, thank you for the suggestion.

PR #166, opened 2021-06-02
Fix YaraFileBuilder recheck with custom modules

TadeasKucera:
This PR fixes a problem of `YaraFileBuilder::get` method, which when given `recheck=True` did not use any custom provided modules in the `YaraFileBuilder::_module_pool` attribute of the particular `YaraFileBuilder` instance.

- [x] Add failing test reproducing the problem
- [x] Fix the problem


TadeasKucera (commented):
As the test shows, the PR solves precisely the problem which occured.

PR #168, opened 2021-06-03
Add two tests showing how to steal condition or delete metas

TadeasKucera:


metthal:
Should we close this or does it serve any purpose?

PR #170, opened 2021-06-04
Hotfix for user-defined arrays to be handled correctly by autoformatter

metthal:
It's still not ideal because for example for rule

```yara
	condition:
		1 of [
			cuckoo.sync.mutex(/a/),
			cuckoo.sync.mutex(/b/)
		]
```

is formatted into

```yara
	condition:
		1 of [
		cuckoo.sync.mutex(/a/),
		cuckoo.sync.mutex(/b/)
		]
}
```

but at least it no longer crashes. Once it's merged please release 3.9.2 as a hotfix release.


TadeasKucera (approved):
Thank you very much for such fast hotfix!

Issue #177, opened 2021-06-30
Fix wrong Float representation and parsing


TadeasKucera:
Float values are not parsed / stored correctly.

- Longer floats are trimmed: 0.869932552472 becomes 0.869933, which is undesirable.
- Many zeros behind the dot, like 0.0000000001 are stored as 1e-10 which is also not wanted.

I have added two tests into a new branch [fix_float_values](https://github.com/avast/yaramod/tree/fix_float_values) which demostrate the problem.

metthal:
Resolved in #179

PR #179, opened 2021-07-16
Double always with decimal 

xbabka01:
Hi, this is fix of bug which ocured when you wanted to create `double` value without decimal (example 10.0). Yaramod in this case generated int instead of `double` in result

Before
```python
yaramod.double_val(10.0).get().get_text() == '10'
```
now
```python
yaramod.double_val(10.0).get().get_text() == '10.0'
```

### DIFF
--- a/include/yaramod/utils/utils.h
+++ b/include/yaramod/utils/utils.h
@@ -9,6 +9,7 @@
 #include <cctype>
 #include <sstream>
 #include <string>
+#include <cmath>
 
 namespace yaramod {
 
@@ -45,15 +46,18 @@ bool startsWith(const std::string& str, const T& withWhat)
  * @return Number converted to string.
  */
 template <typename T>
-std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false, bool toUpper = false)
+std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false, bool toUpper = false, bool showpoint=false)
 {
 	std::ostringstream os;
+	os.precision(std::numeric_limits<double>::max_digits10);
 	if (toUpper)
 		os << std::uppercase;
 	if (showbase)
 		os << format << std::showbase << num;
 	else
 		os << format << num;
+	if (showpoint && (std::floor(num) == num))
+		os << ".0";
 	return os.str();
 }
 
--- a/include/yaramod/utils/utils.h
+++ b/include/yaramod/utils/utils.h
@@ -46,7 +46,7 @@ bool startsWith(const std::string& str, const T& withWhat)
  * @return Number converted to string.
  */
 template <typename T>
-std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false, bool toUpper = false, bool showpoint=false)
+std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false, bool toUpper = false)
 {
 	std::ostringstream os;
 	if (toUpper)
@@ -55,7 +55,7 @@ std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std
 		os << format << std::showbase << num;
 	else
 		os << format << num;
-	if (showpoint && (std::floor(num) == num))
+	if (std::is_floating_point<T>::value && (std::floor(num) == num))
 		os << ".0";
 	return os.str();
 }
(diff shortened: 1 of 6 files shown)


include/yaramod/utils/utils.h
```diff
@@ -45,7 +46,7 @@ bool startsWith(const std::string& str, const T& withWhat)
  * @return Number converted to string.
```
MatejKastak, line 46: Nitpick: Missing docstring with the new argument

src/builder/yara_expression_builder.cpp
```diff
@@ -716,7 +716,7 @@ YaraExpressionBuilder hexIntVal(std::uint64_t value)
 YaraExpressionBuilder doubleVal(double value)
 {
 	auto ts = std::make_shared<TokenStream>();
-	TokenIt token = ts->emplace_back(TokenType::DOUBLE, value, numToStr(value));
+	TokenIt token = ts->emplace_back(TokenType::DOUBLE, value, numToStr(value, std::dec, false, false, true));
```
MatejKastak, line 719: This feels a little cryptic for me.

- which of this flags is relevant for this call?
- which of this is just to satisfy defaults?

However, using this implementation I don't know how to make this better.

MatejKastak (commented):
I was playing a bit more with the `numToStr` function and how to make it work without adding a new argument. What do you think about this approach? I don't think there are cases when you want to have `numToStr(0.0) == "0"`, you should not get the expected value only if you pass in some hidden flag. Considering the backward compatibility I think this should fall into "fix" category?

```c++
template <typename T>
std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false, bool toUpper = false)
{
	std::ostringstream os;

	os << format;

	if (toUpper)
		os << std::uppercase;

	if (showbase)
		os << std::showbase;

	os << num;

	if (std::is_floating_point<T>::value && std::floor(num) == num)
		os << ".0";

	return os.str();
}
```

After some tests I wonder what we should do in cases when the precision is not sufficient. By default the max precision is 6 decimal spaces, according to docs. But if you want to create a floating point number `0.1234567` the last number is truncated in the resulting rule. Maybe its out of the scope for this PR.

# Wrapping up.
Sadly, I have no explicit approval for this PR. I am still curious what do others and you think about the proposed changes.

include/yaramod/utils/utils.h
```diff
@@ -45,7 +46,7 @@ bool startsWith(const std::string& str, const T& withWhat)
...
  */
 template <typename T>
-std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false, bool toUpper = false)
+std::string numToStr(const T num, std::ios_base &(*format)(std::ios_base&) = std::dec, bool showbase = false, bool toUpper = false, bool showpoint=false)
```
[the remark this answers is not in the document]
xbabka01: Removed flag :)

include/yaramod/utils/utils.h
xbabka01: Removed parameter

src/builder/yara_expression_builder.cpp
xbabka01: Yeah, it is. I changed `numToStr` as you suggested to automatically detect floating type instead of using flag

xbabka01:
> After some tests I wonder what we should do in cases when the precision is not sufficient. By default the max precision is 6 decimal spaces, according to docs. But if you want to create a floating point number 0.1234567 the last number is truncated in the resulting rule. Maybe its out of the scope for this PR.

Yeah I saw that  issue #177 and tried to fix it but it broken multiple other tests see [this](https://github.com/avast/yaramod/runs/3089302993). From that I came to the conclusion that it would be better to discuss with what precision yaramod should work

xbabka01:
Precision if fixed in commit e580ffc (#177)

include/yaramod/utils/utils.h
```diff
@@ -48,12 +49,36 @@ template <typename T>
...
+		}
+		if (comma == value.length() - 1)
+		{
+			return value.append("0");
+		}
+		if (comma == value.length() - 2)
+		{
+			return value;
+		}
+		auto zero = value.find_last_not_of('0');
```
metthal, line 79: Could you please post some comments here of what you are trying to achieve because it's not obvious from the get go.

metthal (changes requested):

include/yaramod/utils/utils.h
xbabka01: There are few cases:
* `comma == std::string::npos`: no comma in string so append '.0`
* `comma == value.length() - 1`: Comma is last character so append 0
* `comma == value.length() - 2`: Value have one decimal place, so don't change value
* `else`: trim zeros from end of string presentation of double.  So instead of 10.00000000000000 function will return 10.0

metthal: Okay and why can't we have it in the code also to state the intent? Please put it there. Thanks.

xbabka01: Added comments.

metthal (approved):
LGTM

PR #182, opened 2021-08-05
Added iequals operator

metthal:


TadeasKucera (approved):
LGTM, everything seems to be alright and the python bindings were also added. The only thing that we should consider is to add some documentation, possibly to `docs/rtd/creating_rulesets.rst`. Do you want me to do it?

metthal:
Added to the documentation.

PR #184, opened 2021-08-20
Interchangeable strings and variables sections

tomaskender:
Up until now, you could only define first strings and then variables. Defining variables and then strings was not supported by the parser, even though YARA supports such rule.

The only supported way of defining strings and variables:
```
rule r {
  strings:
    $1 = "one"
  variables:
    var = 123
  condition:
    true
}
```
Unsupported rule that throws an error:
```
rule r {
  variables:
    var = 123
  strings:
    $1 = "one"
  condition:
    true
}
```

From now on, the parser will successfully parse rules with string and variable sections defined in reverse order as well. The default section order when regenerating such rule back into fulltext representation is going to stay like it used to- first strings and then variables.

### DIFF
--- a/include/yaramod/parser/value.h
+++ b/include/yaramod/parser/value.h
@@ -14,6 +14,7 @@
 #include "yaramod/types/rule.h"
 #include "yaramod/types/regexp.h"
 #include "yaramod/types/token_stream.h"
+#include "yaramod/types/sections_summary.h"
 
 #pragma once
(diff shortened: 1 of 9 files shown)


TadeasKucera:
Hello, could you please add a test for the new feature?

TadeasKucera (approved):
I went through the code and found no issues, good job!

Issue #149, opened 2021-01-03
Formatted text should strip the trailing whitespace from the comments


MatejKastak:
Consider a following example:

```python
import yaramod

ymod = yaramod.Yaramod()
f = ymod.parse_string(
    r"""
rule asd {
    // Comment with a trailing space 
    // Comment with a trailing tab	
    // Comment with a trailing combination 	 
    condition:
        false
}
"""
)

print(f.text_formatted)
print(repr(f.text_formatted))
```

All comments contain trailing whitespace and here is anotated the output:

```
'\nrule asd\n{\n\t\t// Comment with a trailing space \n\t\t// Comment with a trailing tab\t\n\t\t// Comment with a trailing combination \t \n\tcondition:\n\t\tfalse\n}\n'
                                                    ^                                    ^                                             ^
                                                    |                                    |                                             +-- space, tab, space
                                                    |                                    +-- tab
                                                    +-- single space character

```

Just to make sure the is there same file in case it is not clear where the whitespace is from the code preview [test.py.zip](https://github.com/avast/yaramod/files/5761621/test.py.zip).

I can't think of a case where keeping the trailing whitespace is intended, especially in the "formatted_text".

PR #189, opened 2021-11-11
Trim whitespaces of oneline comments

TadeasKucera:
Solving #149. Please note the following properties of the solution:

- Only one-line comments are stripped of trailing whitespaces, because when somebody prefers /* comment */ to /*comment*/ I do not intend to make them angry.
- The comments are stripped during their construction, thus only once. I could have made it the other way, so that the comments would be stripped every time we print TokenStream, but we do not need to store the information about where were the whitespaces. Sorry for mentioning this if it is obvious.


MatejKastak (approved):
LGTM :+1: Good job

PR #191, opened 2021-11-28
Fix escaped character handling in regex

### LINKED ISSUE #186: Escaped characters with operators are parsed incorrectly
I've come across an issue where hex-encoded characters with operators are processed incorrectly.
For example, `\x31+`: instead of applying the operator to `\x31` it's incorrectly applied only to `1` and the rest of the character is converted to `RegexpText`.

Installation source: master (`1acefa3`)
POC:

```python
import yaramod

y = yaramod.Yaramod(yaramod.Features.AllCurrent)
yara_file = y.parse_string(r'''
rule regex_test {
    strings:
        $a = /1\x32?3/
    condition:
        $a
}''')
for rule in yara_file.rules:
    for string in rule.strings:
        for unit in string.unit.units:
            print(type(unit), unit.text)
```

```
<class 'yaramod.RegexpText'> 1
<class 'yaramod.RegexpText'> \x
<class 'yaramod.RegexpText'> 3
<class 'yaramod.RegexpOptional'> 2?
<class 'yaramod.RegexpText'> 3
```

msm-code:
I've tried to fix issue #186.

This was suspiciously easy (especially compared to my previous PR: #93), so please check that I don't do something obviously wrong. FWIW this fixes the issue at hand. Using the POC from the linked issue:

```python
import yaramod

y = yaramod.Yaramod(yaramod.Features.AllCurrent)
yara_file = y.parse_string(r'''
rule regex_test {
    strings:
        $a = /1\x32?3/
    condition:
        $a
}''')
for rule in yara_file.rules:
    for string in rule.strings:
        for unit in string.unit.units:
            print(type(unit), unit.text)
```

Previous results:

```
<class 'yaramod.RegexpText'> 1
<class 'yaramod.RegexpText'> \x
<class 'yaramod.RegexpText'> 3
<class 'yaramod.RegexpOptional'> 2?
<class 'yaramod.RegexpText'> 3
```

Current (correct) results:
```
<class 'yaramod.RegexpText'> 1
<class 'yaramod.RegexpOptional'> \x32?
<class 'yaramod.RegexpText'> 3
```

I've also added a test.

Fixes #186.

### DIFF
--- a/src/parser/parser_driver.cpp
+++ b/src/parser/parser_driver.cpp
@@ -375,6 +375,9 @@ void ParserDriver::defineTokens()
 
 		return std::make_pair(range, range);
 	});
+	_parser.token(R"(\\x[a-zA-Z0-9]{2})").states("$regexp").symbol("REGEXP_ESCAPE").description("regexp escaped character").action([](std::string_view str) -> Value {
+		return std::string{str};
+	});
 	_parser.token(R"([^\\\[\(\)\|\$\.\^\+\+*\?])").states("$regexp").symbol("REGEXP_CHAR").description("regexp character").action([](std::string_view str) -> Value {
 		return std::string{str};
 	});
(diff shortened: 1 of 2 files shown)


src/parser/parser_driver.cpp
```diff
@@ -375,6 +375,9 @@ void ParserDriver::defineTokens()
 
 		return std::make_pair(range, range);
 	});
+	_parser.token(R"(\\x[a-zA-Z0-9]{2})").states("$regexp").symbol("REGEXP_ESCAPE").description("regexp escaped character").action([](std::string_view str) -> Value {
```
msm-code, line 378: Matches hex escaped characters, like  "\x61", as a single token.

nazywam: Should we care about other encoding methods as well? It seems like `\65?` is a valid regex as well, not sure about the YARA side though 🤔

msm-code: I *think* yara doesn't support that:
```
rule kot {
    strings: $a = /kot\61kot/
    condition: $a
}
```
```
❯ yara -rs kot.yar /tmp/a.txt
kot.yar(2): error: invalid regular expression "$a": backreferences are not allowed
```
But sure, we should check for any other alternative encodings (if they exist). By the way I've checked, and `\X61` is not a valid encoding (`x` must be lowercase).

TadeasKucera:
Thank you very much for this very nicely written PR. It is good, I will merge it.

Issue #141, opened 2020-09-10
Empty `strings` part of the rule is parsed without error


stepanek-m:
When strings keyword is present in the rule, but there are no strings, YARA fails with error. However, yaramod parses it without error.
Example:
```
rule a {
    strings:
    condition:
        true
}
```

PR #192, opened 2021-12-01
Make empty strings section invalid

msm-code:
Another small PR from me. This time I've tackled #141.

After merging this, empty `strings:` section won't be allowed. I achieve this by adding a new rule for `strings_body_nonempty` that cannot be empty, and using it instead of `strings_body` in the main derivation. 

This is not related to any issue with mquery, but saw that issue on the issue list and figured I can try to solve it. I realise it's very low priority, but probably also won't hurt.

fixes: #141

### DIFF
--- a/src/parser/parser_driver.cpp
+++ b/src/parser/parser_driver.cpp
@@ -652,13 +652,13 @@ void ParserDriver::defineGrammar()
 	}
 
 	_parser.rule("strings") // shared_ptr<StringsTrie>
-		.production("STRINGS", "COLON", "strings_body", [](auto&& args) -> Value {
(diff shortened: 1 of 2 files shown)


TadeasKucera (approved):
Thank you for this PR, it does precisely what your description says. By adding one non-terminal symbol to the grammar you postpone the possibility to generate empty word from the non-terminal `strings_body_nonempty`. It is correct and I will merge it. I will also make some cosmetics changes immediately after merge, because we are used to put the semicolon on a new line in the grammar definition. Thank you very much!

msm-code:
Glad i could help!
Thanks for mentioning the semicolon formatting - I'll keep that in mind in the future.

PR #196, opened 2022-01-28
Add dynsym_entries and dynsym ELF module attributes

HoundThe:
This PR includes missing attributes from YARA ELF module - dynsym_entries and dynsym


mienkofax (approved):
LGTM :+1:.

PR #198, opened 2022-02-09
YARA-1554: percentage support

TadeasKucera:
Corresponding changes in YARA:
https://github.com/VirusTotal/yara/commit/32c969f1e4ceed1f3b4df03f29851d8ea75735c6


include/yaramod/types/expressions.h
```diff
@@ -400,23 +409,36 @@ class UnaryOpExpression : public Expression
 
 protected:
 	template <typename ExpPtr>
-	UnaryOpExpression(TokenIt op, ExpPtr&& expr)
+	UnaryOpExpression(TokenIt op, ExpPtr&& expr, bool operatorLeft)
```
metthal, line 412: Over the years I kinda refrained from boolean parameters/arguments to more strictly typed values as it provides more explainability and type security. So instead of having `bool operatorLeft` I'd welcome something in lines of

```c++
enum class UnaryOperatorPlacement
{
    Left,
    Right
}
```

metthal: Also if you want to push this idea even further, you can have

```c++
template <typename ExpPtr, UnaryOperatorPlacement OpPlacement>
class UnaryOpExpression { ... }
```

and then you can directly inherit from `UnaryOpExpression<_, UnaryOpPlacement::Left>` or `UnaryOpExpression<_, UnaryOpPlacement::Right>`. You'll avoid having if's in your `getText` and you can still have very simple constructor.

src/parser/parser_driver.cpp
```diff
@@ -1278,6 +1278,27 @@ void ParserDriver::defineGrammar()
 			output->setTokenStream(currentTokenStream());
 			return output;
 		})
+		.production("primary_expression", "PERCENT", "OF", "string_set", [&](auto&& args) -> Value {
+			auto for_expr = std::move(args[0].getExpression());
+			TokenIt percent = args[1].getTokenIt();
+			size_t value = 0;
```
metthal, line 1284: Any specific reason for `size_t` which is usually used when we are dealing with memory sizes and not something like `uint8_t` since we are storing just values between 1 and 100?

metthal (changes requested):
What I'm missing is some sort of builder function to build me this new percentage expression (+ builder test). Otherwise, just small nitpicks.

include/yaramod/types/expressions.h
TadeasKucera: Thank you very much for the tip. I tried to push it further, so I introduced
```
template <typename ExpPtr, UnaryOperatorPlacement OpPlacement>
class UnaryOpExpression : public Expression
```
but I don't know how to create two virtual methods `getText` for both options and I also struggle with the inheritance:
```
error: ‘_’ was not declared in this scope:
class DefinedExpression : public UnaryOpExpression<_, UnaryOperatorPlacement::Left>
```
I will try more tomorrow.

metthal: Right, in that case, don't use the second approach. I forgot that it's all implemented in virtual method.

TadeasKucera: Done, thank you.

TadeasKucera: Done, thank you!

TadeasKucera:
Thank you very much for your early coming review, I added the builder and addressed all the issues that you listed as you can see above.

src/parser/parser_driver.cpp
TadeasKucera: I see, thank you, I fixed it :+1:

metthal (approved):

PR #199, opened 2022-02-10
YARA-1550: Yara42 new math attributes

TadeasKucera:
Mirrors the changes in yara introduced here:
https://github.com/VirusTotal/yara/commit/87bc03d58743aaaa0f11b6d19d1d26e717f3cb01
https://github.com/VirusTotal/yara/commit/0d5f2cd5ef2595d93886db05d970107f83c494d7
https://github.com/VirusTotal/yara/commit/2edd10c735d91120f0f567c893093c873a67ba95
Solves YARA-1550, YARA-1552 and YARA-1556.

### DIFF
--- a/tests/cpp/parser_tests.cpp
+++ b/tests/cpp/parser_tests.cpp
@@ -3513,7 +3513,7 @@ rule magic_module
 }
 
 TEST_F(ParserTests,
-MathModuleWorks) {
+MathModuleWorks1) {
 	prepareInput(
 R"(
 import "math"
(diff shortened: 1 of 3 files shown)


tests/cpp/parser_tests.cpp
```diff
@@ -3521,17 +3521,17 @@ import "math"
 rule math_module
 {
 	condition:
-		math.entropy("dummy") > 7
+		math.to_number(math.entropy("dummy") > 7) == 1
```
metthal, line 3524: Should this be a brand new test instead of rewriting existing one?

TadeasKucera: We do not have a test for each module attribute, so I personally think there is no trouble with testing more functions within one test. But I can create a new test if you would prefer it.

metthal: Yeah, just put there more attributes.

TadeasKucera: Done :+1:

metthal (approved):

PR #200, opened 2022-02-10
YARA-1551: Yara42 pe module attributes

TadeasKucera:
Additionally to the ticket's instructions, we migrate:
pe.imports and pe.number_of_delayed_imports
We also solve YARA-1747 by adding pe.entry_point_raw


MatejKastak (approved):
Also LGTM :+1:

PR #204, opened 2022-02-15
YARA-1553: Add operator in

TadeasKucera:
This PR allows Yaramod to parse and accept conditions similar to these:
`all of ($a*) in (filesize-500..filesize)` introduced to YARA here:
https://github.com/VirusTotal/yara/commit/24a5ad1383aa5df0f42f9ef7c6d77eba8afd2cf9
`#a in (filesize-500..filesize) == 2` introduced to YARA here:
https://github.com/VirusTotal/yara/commit/e71801644f73771d8003c14fa3c5237d22319fd0


include/yaramod/types/expressions.h
```diff
@@ -1209,6 +1209,50 @@ class OfExpression : public ForExpression
...
+/**
+ * Class representing 'of' expression. 'of' expression is shortened version of
+ * for loop over string set with no loop body. It has hidden body which always contains just ( $ ).
+ *
+ * For example:
+ * @code
+ * all of ($str1, $str2) in (filesize-1000, filesize)
+ * @endcode
+ */
+class OfInRangeExpression : public ForExpression
```
metthal, line 1221: What I wonder about is whether we really need a completely new type of an expression or whether it wouldn't be better to have it as an optional part of `OfExpression`. Have you though about this option? What are the pros/cons compared to this?

metthal (changes requested):

include/yaramod/types/expressions.h
TadeasKucera: That is a good idea, the pros would be, that it requires less code to be written (which was already written) but also less code to be maintained. I do not really see any cons, I can try to implement it in next commit so we can decide when we see the code.

src/builder/yara_expression_builder.cpp
```diff
@@ -1099,6 +1099,29 @@ YaraExpressionBuilder of(const YaraExpressionBuilder& ofExpr, const YaraExpressi
...
+/**
+ * Creates the expression with match of the specified set of strings in given range.
+ *
+ * @param quantifier All / Any / None expression.
+ * @param set Set expression.
+ * @param range Range expression.
+ *
+ * @return Builder.
+ */
+YaraExpressionBuilder ofInRange(const YaraExpressionBuilder& quantifier, const YaraExpressionBuilder& set, const YaraExpressionBuilder& range)
```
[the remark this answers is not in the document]
TadeasKucera: I renamed it, it is much better I think. Thank you for the idea! :+1:

metthal (approved):

metthal:
Could you please resolve the conflict? You can merge it afterwards.

PR #206, opened 2022-03-10
YARA-1806: Fix removing of parts of rules in tokenstream

TadeasKucera:
Removing all relevant tokens when deleting YARA rule from `YaraFile` instance.

### DIFF
--- a/src/types/rule.cpp
+++ b/src/types/rule.cpp
@@ -276,6 +276,10 @@ const Meta* Rule::getMetaWithName(const std::string& key) const
  */
 TokenIt Rule::getFirstTokenIt() const
 {
+	if (_mod_private.has_value())
+		return _mod_private.value();
+	if (_mod_global.has_value())
+		return _mod_global.value();
 	return _tokenStream->findBackwards(TokenType::RULE, _name);
 }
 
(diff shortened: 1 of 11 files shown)


src/types/rule.cpp
```diff
@@ -276,6 +276,10 @@ const Meta* Rule::getMetaWithName(const std::string& key) const
  */
 TokenIt Rule::getFirstTokenIt() const
 {
+	if (_mod_private.has_value())
```
metthal, line 279: Wouldn't this assume that there is a specific order in which rule modifiers are specified? But YARA grammar allows to specify both `private global` and `global private`.

TadeasKucera: Yes, I assumed that `private` comes before `global`, which is a mistake, thank you very much for noticing! Will fix that.

src/types/rule.cpp
```diff
@@ -276,6 +276,30 @@ const Meta* Rule::getMetaWithName(const std::string& key) const
  */
 TokenIt Rule::getFirstTokenIt() const
 {
+	if (isPrivate() && isGlobal())
```
metthal, line 279: This feels like a workaround for the problem. What if the new modifier is added? Are we going to list out all the possible combinations? Why not storing the modifiers in some sort of containers. Whenever the first token is requested just the first token is returned from the container.

metthal (changes requested):

src/types/rule.cpp
TadeasKucera: Thank you for pointing this out. I have changed the way we work with modifiers so it should be much easier now to add more rule modifiers. Please see the last commit.

PR #209, opened 2022-03-17
Processes

tomaskender:
This new feature allows users to define attributes that reference an existing structure.

Use case:
```
Person - name <string>
       - partner <Person>
       - children <Person[]>
 
 person.parents[0].parents[0].name
```

Changes had to be made to *_addStruct* and other parts of *modules.cpp*. The most serious one was the predeclaration of a struct symbol in it's parent object. It's assigned as an attribute of it's parent **before** attributes of the struct itself are loaded.

Sample declaration of references in modules:
```
module: my_module
{
   "kind": "array",
  "name": "people",
  "structure":
  {
    "kind": "struct",
    "name": "people",
    "attributes": [
      {
        "kind": "value",
        "name": "name",
        "type": "s"
      },
      {
        "kind": "reference",
        "name": "partner",
        "type": "my_module.people",
      },
      {
        "kind": "array",
        "name": "children",
        "structure":
        {
          "kind": "reference",
          "name": "children",
          "type": "my_module.people"
        }
      }
    ]
  }
}
```

### DIFF
--- a/src/types/modules/module.cpp
+++ b/src/types/modules/module.cpp
@@ -37,6 +37,28 @@ std::optional<ExpressionType> stringToExpressionType (const std::string& str)
 }
 
 /**
+ * Finds an attribute with specified name for a structure or iterable base.
+ */
+std::optional<std::shared_ptr<Symbol>> _getExistingAttribute(Symbol* base, std::string& name)
+{
+	if (base)
+	{
+		switch (base->getType())
+		{
+			case Symbol::Type::Structure:
+				return ((StructureSymbol*)base)->getAttribute(name);
+			case Symbol::Type::Array:
+			case Symbol::Type::Dictionary:
+			{
+				auto structElement = ((IterableSymbol*)base)->getStructuredElementType();
+				return structElement ? std::make_optional(structElement) : std::nullopt;
+			}
+		}
+	}
+	return std::nullopt;
+}
+
+/**
  * Constructor.
  *
  * @param name Name of the module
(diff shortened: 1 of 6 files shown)


src/types/modules/module.cpp
```diff
@@ -36,6 +36,30 @@ std::optional<ExpressionType> stringToExpressionType (const std::string& str)
...
+	if (base)
+	{
+		switch (base->getType())
+		{
+			case Symbol::Type::Structure:
+				return ((StructureSymbol*)base)->getAttribute(name);
+			case Symbol::Type::Array:
+			case Symbol::Type::Dictionary:
+			{
+				auto structElement = ((IterableSymbol*)base)->getStructuredElementType();
```
TadeasKucera, line 53: Casting base to derived class is always risky even when you check that you deal with the particular derived class instance. The best is to avoid such needs with different design when possible and when in real need it is better to use dynamic cast here. Is dynamic cast appliable here? Can we use something safer than (<>*) cast?

src/types/modules/module.cpp
```diff
@@ -403,4 +496,81 @@ void Module::_importJson(const Json& json)
...
+ * Adds an attribute to it's base that can be either a parent structure or an iterable.
+ */
+void Module::_addObjectToBase(Symbol* base, std::shared_ptr<Symbol> newAttribute)
+{
+	if (base)
+	{
+		switch (base->getType())
+		{
+			case Symbol::Type::Structure:
+				((StructureSymbol*)base)->addAttribute(newAttribute);
```
TadeasKucera, line 564: Again quite a risky cast. In Yaramod we use `static_pointer_cast` to cast from pointer to derived class to pointer to base class which is safe, but not appliable here. If I am not mistaken, so far we did not need to cast base class to derived class. Is there any way we would avoid it here?

TadeasKucera, line 564: Again quite a risky cast. In Yaramod we use `static_pointer_cast` to cast from pointer to derived class to pointer to base class which is safe, but not appliable here. If I am not mistaken, so far we did not need to cast base class to derived class. Is there any way we would avoid it here?

TadeasKucera (changes requested):
The code is nicely formatted and after going through it I believe it is fully functional. I am a bit concerned with the usage of cast from base class to a derived class as you can read above. Please elaborate on that and tell me what you think. Otherwise it seems very good, thank you.

src/types/modules/module.cpp
tomaskender: We need access to the derived class because we want to be able to get and set the attribute of an iterable or structure base. I've tried to think of a way to avoid conversion, but that would lead to a duplication of *_addStruct()*- there would be one for a structure base and the other for an iterable base.
That would make it difficult to maintain or extend the code, should we attempt similar in-depth changes to the modules like I did in this PR, since functions in this module are usually called in a complicated circular dependency manner and further splitting of functionality would only make it worse.
Thank you for pointing out the possibility of using a dynamic_cast, I've adjusted the code to reflect that.

src/types/modules/module.cpp
tomaskender: As explained above, the cast is needed, because iterables and structures use different functions to get and set their attributes. I've changed the cast here to the same dynamic cast as in the other example, let me know if it's okay

TadeasKucera (approved):
Thank you for the changes, I think it is LGTM. Good job!

PR #213, opened 2022-05-12
Added support for N of <rule_set> expressions

metthal:
YARA recently added support for `<N> of <rule_set>` expression where you can also express rules as wildcards.

Since we can't distinguish between `<ID> <*>` in tokenizer because of the possible ambiguity with multiplication, we need to handle it in the grammar. The way I have dealt with it is that we expect 2 tokens - `<ID> <MULTIPLY>` and then manually turn them into `<ID_WILDCARD>` tokens.

I've also added exception for these tokens into tokenstream so that 2 successive `ID_WILDCARD` tokens are put into tokenstream without space in between them.

In order to quickly evaluate the prefixes specified by these wildcards, I've also added storing of `Rule` points into trie by their name.

Some other things which I encountered and fixed in this PR:

* Builder incorrectly marked parentheses when constructing `set()` as `LP` and `RP` and not `LP_ENUMERATION` and `RP_ENUMERATION`, so the special handling for in enumeration formatting didn't work like expected.
* Dump rules AST example code didn't contain all the new types of expressions so I've added them
* Simplified grammar rules for `<N> of <string_set>` by introducing `string_enumeration_item` nonterminal, so that we have less repeated code.
* `OfExpression` didn't have `range` exposed in Python bindings after `<N> of <set> in <range>` has been introduced so I've added it


TadeasKucera (approved):
Thank you very much for adding this. It looks all good to me.

PR #217, opened 2022-06-05
Support for removal of string modifiers

tomaskender:
* Allow resetting the list of modifiers for each string
* Introduce read-only access to raw modifier list


MatejKastak (approved):
LGTM :+1:

I think it would be also nice to include test for `getModifiersText`. 

```python

    def test_get_string_modifiers(self):
        yara_file = yaramod.Yaramod().parse_string(
            """
rule rule_with_string_modifiers {
	strings:
		$00 = "value" ascii wide fullword
	condition:
		all of them
}"""
        )

        self.assertEqual(len(yara_file.rules), 1)
        self.assertEqual(len(yara_file.rules[0].strings[0].modifiers), 3)

        # Order of modifiers is not guaranteed
        string = yara_file.rules[0].strings[0]
        expected_modifiers = {
            yaramod.FullwordStringModifier,
            yaramod.WideStringModifier,
            yaramod.AsciiStringModifier,
        }
        for modifier in string.modifiers:
            assert type(modifier) in expected_modifiers
            expected_modifiers.remove(type(modifier))
```

include/yaramod/types/string.h
```diff
@@ -245,6 +253,17 @@ class String
 		_mods[modifier->getType()] = modifier;
 		return true;
 	}
+
+	void removeModifiers()
+	{
+		for (const auto& mod : _mods)
```
metthal, line 259: Just a heads up for the future, there are now [structured bindings](https://en.cppreference.com/w/cpp/language/structured_binding) since C++17 which make working with pairs/tuples much better.

metthal (approved):

PR #218, opened 2022-06-07
Generate UIDs for AST nodes with Expression base class

HoundThe:
I've added UIDGenerator as part of the ParserDriver, which assigns UIDs to AST nodes that have `class Expression` as their base class.

The UIDs are supposed to be unique for a given input and deterministic, they are implemented using a 64bit resettable counter. The counter is reset when new parsing input is provided using the existing `prepareParser()` function and when the condition rule is applied, so it's reset before parsing each condition.

I've modified the existing rule condition AST dumper example code, to print the UIDs and for input:
```
rule ExampleRule
{
    strings:
        $my_text_string = "text here"
        $my_hex_string = { E2 34 A1 C8 23 FB }
        $a = "text1"
    condition:
        $my_text_string or $my_hex_string and
		not $a or false
}

rule ExampleRule1
{
    strings:
        $my_text_string = "text here" private
        $my_hex_string = { E2 34 A1 C8 23 FB }
        $a = "text1"
        $b = "text2"
        $c = "text3"
        $d = "text4"
    condition:
        ($a or $b) and ($c or $d)
        or (#a == 6 and #b > 10) and
        any of them
}

```
I get the following output:

```
==== RULE: ExampleRule
Or[0x55c587337c60]  uid=7
    Or[0x55c587335840]  uid=5
        String[0x55c587318880]  id=$my_text_string uid=0
        And[0x55c5873357d0]  uid=4
            String[0x55c587334230]  id=$my_hex_string uid=1
            Not[0x55c587332890]  uid=3
                String[0x55c587335720]  id=$a uid=2
    BoolLiteral[0x55c587337660]  value=false uid=6
==== RULE: ExampleRule1
Or[0x55c587346980]  uid=21
    And[0x55c5873414a0]  uid=8
        Parentheses[0x55c587340bb0]  uid=3
            Or[0x55c587340590]  uid=2
                String[0x55c58731b910]  id=$a uid=0
                String[0x55c587340540]  id=$b uid=1
        Parentheses[0x55c587341430]  uid=7
            Or[0x55c5873411c0]  uid=6
                String[0x55c587340f20]  id=$c uid=4
                String[0x55c587341170]  id=$d uid=5
    And[0x55c587346910]  uid=20
        Parentheses[0x55c587344c20]  uid=16
            And[0x55c587344ab0]  uid=15
                Equal[0x55c587343b40]  uid=11
                    StringCount[0x55c587342f20]  id=#a uid=9
                    IntLiteral[0x55c587343840]  value=6 uid=10
                GreaterThan[0x55c587344a40]  uid=14
                    StringCount[0x55c5873441b0]  id=#b uid=12
                    IntLiteral[0x55c5873449f0]  value=10 uid=13
        Of[0x55c587346870]  uid=19
            Any[0x55c5873464b0]  uid=17
            Them[0x55c587346820]  uid=18
 ```

### DIFF
--- a/include/yaramod/types/expression.h
+++ b/include/yaramod/types/expression.h
@@ -57,7 +57,7 @@ public:
 
 	/// @name Getter methods
 	/// @{
-	uint64_t getUid() const { return _uid; }
+	std::uint64_t getUid() const { return _uid; }
 	Expression::Type getType() const { return _type; }
 	std::string getTypeString() const
 	{
@@ -88,7 +88,7 @@ public:
 
 	/// @name Setter methods
 	/// @{
-	void setUid(uint64_t uid) { _uid = uid; }
+	void setUid(std::uint64_t uid) { _uid = uid; }
 	void setType(Expression::Type type) { _type = type; }
 	void setTokenStream(const std::shared_ptr<TokenStream>& ts) { _tokenStream = ts; }
 	/// @}
@@ -160,7 +160,7 @@ protected:
 
 private:
 	Type _type; ///< Type of the expression
-	uint64_t _uid;
+	std::uint64_t _uid;
 };
 
 }
--- a/include/yaramod/types/expression.h
+++ b/include/yaramod/types/expression.h
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <memory>
 #include <sstream>
 #include <string>
@@ -56,6 +57,7 @@ public:
 
 	/// @name Getter methods
 	/// @{
+	uint64_t getUid() const { return _uid; }
 	Expression::Type getType() const { return _type; }
 	std::string getTypeString() const
 	{
@@ -86,6 +88,7 @@ public:
 
 	/// @name Setter methods
 	/// @{
+	void setUid(uint64_t uid) { _uid = uid; }
 	void setType(Expression::Type type) { _type = type; }
 	void setTokenStream(const std::shared_ptr<TokenStream>& ts) { _tokenStream = ts; }
 	/// @}
@@ -157,6 +160,7 @@ protected:
 
 private:
 	Type _type; ///< Type of the expression
+	uint64_t _uid;
 };
 
 }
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/python/yaramod_python.cpp [hunk 1]
```diff
@@ -537,6 +538,7 @@ void addExpressionClasses(py::module& module)
 	py::class_<Expression, std::shared_ptr<Expression>>(module, "Expression")
 		.def("accept", &Expression::accept)
 		.def("get_text", &Expression::getText, py::arg("indent") = std::string{})
+		.def("get_uid", &Expression::getUid)
```
metthal, line 541: Could we maybe change it into property that would automatically default to call of `getUid`? It transform what is rather C++ style of doing things (getter methods) into Pythonic code (getter properties).

include/yaramod/types/expression.h
```diff
@@ -56,6 +57,7 @@ class Expression
 
 	/// @name Getter methods
 	/// @{
+	uint64_t getUid() const { return _uid; }
```
metthal, line 60: Please make it `std::uint64_t` to make it use C++ namespace and also to align it with the rest of the code base.

metthal (changes requested):
Could we please also mention this in a documentation in docs folder which is then getting generated for our ReadTheDocs? Otherwise, looks good and I think this would suffice.

HoundThe:
Thanks a lot for the review! I've incorporated the feedback.

src/python/yaramod_python.cpp
```diff
@@ -538,6 +539,7 @@ void addExpressionClasses(py::module& module)
 		.def("accept", &Expression::accept)
 		.def("get_text", &Expression::getText, py::arg("indent") = std::string{})
 		.def("exchange_tokens", py::overload_cast<Expression*>(&Expression::exchangeTokens))
+		.def_property_readonly("uid", [](const Expression& self) { return self.getUid();} )
```
metthal, line 542: You can directly use `&Expression::getUid`. You have maybe copied it over from the code below but there it's needed because even though we are exposing `text` as a property, the function `getText` has a default argument which Python cannot handle. If you look even more down, you can see that we are many times actually mapping the properties directly to method pointers.

metthal (changes requested):

src/python/yaramod_python.cpp
HoundThe: Oh ye, thanks for that, fixed

PR #233, opened 2023-01-12
ci: Be more explicit when excluding from matrix

MatejKastak:
Not sure why is this necessary. Based on the documentation:

> An excluded configuration only has to be a partial match for it to be
> excluded.

this should work in the original form. And it seems it was working in the
past.


MatejKastak:
This is not necessary, the added tests were caused by the repository settings.

PR #234, opened 2023-01-17
Add TokenStream manipulation methods to Python bindings

Tasssadar:


metthal (changes requested):
Thanks for the contribution. I have no objection against the changes but if you could please add a few tests into `tests/python` to ensure that those methods work as intended? Thank you.

MatejKastak (approved):
Thank you! :+1:

PR #235, opened 2023-01-18
Python type stubs

Tasssadar:
(Requires #234)

This is my attempt to add at least _some_ support for code completion and type checking to yaramod.

The initial `.pyi` file is generated by `stubgen`, but the generated output did not even have correct syntax, so I had to make some manual adjustments and I also added some more types while at it - the PR is split into multiple commits on purpose, so you can see the diff between generated output and what actually works.

I changed the build scripts to include the needed `.pyi` and `py.typed` file. I'm not sure I will not break anything for existing uses of yaramod, especially in projects that already use `mypy`, so this should definitely be at least a minor version bump.


src/python/yaramod_python.cpp
```diff
@@ -171,7 +171,7 @@ void addEnums(py::module& module)
 		.value("Comma", TokenType::COMMA)
 		.value("Private", TokenType::PRIVATE)
 		.value("Global", TokenType::GLOBAL)
-		.value("None", TokenType::NONE)
+		.value("NoneToken", TokenType::NONE)
```
Tasssadar, line 174: This should be safe because there is no real way to use the previous `None` name - it's a keyword, so `TokenType.None` returned error. I suppose `getattr(TokenType, "None")` would work, but is rather cumbersome. 

I intially wanted to keep both `None` and `NoneToken`, but Enums can't have duplicates.

MatejKastak: Correct, thanks for catching this bug! I also believe this should be safe.

Just a reference:
```
>>> import yaramod
>>> yaramod.TokenType.Global
<TokenType.Global: 51>
>>> yaramod.TokenType.None
  File "<stdin>", line 1
    yaramod.TokenType.None
                      ^^^^
SyntaxError: invalid syntax
```

.github/workflows/pipeline.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           ${{ startsWith(matrix.os, 'windows') && format('build\tests\cpp\{0}\yaramod_tests.exe', matrix.config) || 'build/tests/cpp/yaramod_tests' }}
       - name: Python Tests
         run: |
-          python -m pip install pytest
+          python -m pip install pytest mypy==0.991
```
MatejKastak, line 67: Nitpick: Looks like we are using `requirements.txt` in the repo root for dev dependencies.

Can you also specify this dependency there. Dependabot will be able to update that in the future (once we set it up). And we will be less likely to forget about it, that it is required by tests.

MatejKastak (approved):
Very nice improvement! :+1:

Can you shortly summarize how did you create type stubs? I would like to create a wiki page for that how to generate them, how to update them, and optionally how they are tested.

.github/workflows/pipeline.yml
Tasssadar: It looks like it is not used here though, otherwise I don't understand why the `python -m pip install pytest` was there :thinking:

MatejKastak: You are right, the `pytest` in the requirements.txt file is also locked to some ancient version. It will be a good candidate for cleanup, based on the `git blame` it could be an old remnant from Travis CI. Resolved.

metthal (approved):

PR #244, opened 2023-04-18
feat(modules): Add support for string module

MatejKastak:
YARA v4.3.0 added new string module. This commit adds support to the new module.

Ref: https://yara.readthedocs.io/en/stable/modules/string.html


MatejKastak:
Thanks for review, `YARA_SYNTAX_VERSION` version bumped. I will merge after tests pass.

PR #247, opened 2023-06-12
fix(parser): Fixed parsing of empty alternation groups (#193)

### LINKED ISSUE #193: Regular expression idiom (a|b|) accepted by yara but not yaramod
A regular expression with an alternation having the empty string as one possibility, such as `(a|b|)`, produces a parse error in yaramod but is accepted by yara and works as expected. The python code below demonstrates the incompatibility.

As a workaround, the equivalent expression `(a|b)?` is accepted by both yara and yaramod.

```python
import yaramod
import yara

def test_rule(rule_txt, samples):
    try:
        ym = yaramod.Yaramod(yaramod.Features.AllCurrent)
        ym_file = ym.parse_string(rule_txt)
        print('\tyaramod successful parse')
    except yaramod.ParserError as e:
        print(f'\tyaramod failed parse: {str(e)}')

    try:
        rule = yara.compile(source=rule_txt)
        print('\tyara successful parse')
        for sample_txt in samples:
            if not rule.match(data=sample_txt):
                print(f'\tsample {sample} did not match!')
                return
        print('\tall samples matched as expected')
    except yara.Error as e:
        print(f'\tyara failed parse: {str(e)}')


GOOD_RULE = """
rule good {
strings:
    $re = /a(b|c|d)?z/
condition:
    $re
}
"""
BAD_RULE = """
rule bad {
strings:
    $re = /a(b|c|d|)z/
condition:
    $re
}
"""
SAMPLES = ["az", "abz", "acz", "adz"]

print('GOOD_RULE')
test_rule(GOOD_RULE, SAMPLES)
print('BAD_RULE')
test_rule(BAD_RULE, SAMPLES)
```

metthal:
This fixes parsing of regexes like `/(a|b|)/` which are valid YARA but we weren't able to handle empty alternation groups. Fixes #193.

### DIFF
--- a/src/parser/parser_driver.cpp
+++ b/src/parser/parser_driver.cpp
@@ -993,6 +993,12 @@ void ParserDriver::defineGrammar()
 			std::shared_ptr<RegexpUnit> concat = std::make_shared<RegexpConcat>(args[2].getMultipleRegexpUnits());
 			return std::make_shared<RegexpOr>(std::move(arg), std::move(concat));
 		})
+		.production("regexp_or", "REGEXP_OR", [](auto&& args) -> Value {
+			// Special case in which we end alternation with empty string (like /(a|b|)/) which needs to be handled this way
+			// otherwise the grammar produces shift-reduce conflicts
+			std::shared_ptr<RegexpUnit> arg = std::move(args[0].getRegexpUnit());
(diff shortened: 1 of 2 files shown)


MatejKastak (approved):
LGTM :+1:

I also tested and it is also accepted by YARA.

```yara
rule asd {
    strings:
        $1 = /(a|||||)/ 
    condition:
        $1
}
```

PR #248, opened 2023-06-20
fix: Correct the type of `dotnet.guids` to array

MatejKastak:
Docs: https://yara.readthedocs.io/en/stable/modules/dotnet.html?highlight=dotnet#c.guids


mienkofax (approved):
LGTM :+1:.

PR #249, opened 2023-07-17
feat(modules): Update dotnet module

MatejKastak:


vojone:
I checked it against [YARA documentation of dotnet module](https://yara.readthedocs.io/en/stable/modules/dotnet.html) and everything should be updated now. 

However, during the inspection I found some symbols in file [module_dotnet.json](https://github.com/MatejKastak/yaramod/blob/feat-update-dotnet-module/modules/module_dotnet.json), that are not documented in mentioned doc (they have not been probably added in this update):

`number_of_assembly_refs`
`typelib`
`constants`
`number_of_constants`

But they look as they are implemented in [dotnet module](https://github.com/VirusTotal/yara/blob/master/libyara/modules/dotnet/dotnet.c), so just documentation of them is missing. Should I open PR in upstream yara repo and add them to YARA doc?

MatejKastak:
Good catch :+1: Yes, please open PR to YARA extending the docs.

vojone (approved):
LGTM 👍

PR #252, opened 2023-07-24
Of ... at expression

vojone:
Added support for `... of ... at ...` expression, that is supported in YARA 4.3.


metthal (changes requested):
Since we have `of ... in ...` and `of ... at ...` in YARA now and it's hard to distinguish between those 2 when using function overloads, you decided to introduce `ofAt/of_at`. I'm not against it, but I would prefer if we stop using function overloading for `of ... in ...` and instead name it all explicitly, so in the end, we would have `of`, `ofInRange` and `ofAt`. I think that's better than having `of` for both `of ...` and `of ... in ...` and then `ofAt` for `of ... at ...`.

vojone:
Due to review, I renamed the second overload of `of` to `OfInRange`/`of_in_range`. It should be more clear now. I was originally afraid of that changing the name of the method would break existing applications using yaramod.

src/python/typings/__init__.pyi
```diff
@@ -1601,6 +1602,7 @@ def not_wildcard_low(arg0: int) -> YaraHexStringBuilder: ...
 def of(arg0: YaraExpressionBuilder, arg1: YaraExpressionBuilder) -> YaraExpressionBuilder: ...
 @overload
 def of(arg0: YaraExpressionBuilder, arg1: YaraExpressionBuilder, arg2: YaraExpressionBuilder) -> YaraExpressionBuilder: ...
```
metthal, line 1604: This line needs to be changed too.

metthal (changes requested):

metthal (approved):

PR #264, opened 2024-09-24
feat: Added support for expression arrays with [...] and (...)

metthal:


tests/cpp/parser_tests.cpp
```diff
@@ -8517,5 +8517,50 @@ rule module_rule
...
 }
 
+TEST_F(ParserTests,
+ExpressionArrayWorks) {
+	prepareInput(
+R"(
+rule expression_array
+{
+	condition:
+		any of (1, 2, 3)
```
TommYDeeee, line 8527: Just a note as we have discussed this. The fix for this in upstream is already created: https://github.com/VirusTotal/yara-x/pull/208

TommYDeeee (approved):
From a grammar point of view it LGTM

src/parser/parser_driver.cpp
```diff
@@ -1320,52 +1320,72 @@ void ParserDriver::defineGrammar()
 				return output;
 			}
 		)
-		.production("for_expression", "OF", "string_set", [&](auto&& args) -> Value {
+		//.production("for_expression", "OF", "string_set", [&](auto&& args) -> Value {
```
MatejKastak, line 1323: Nit: What is the reason behind commented out code? Do we want to leave it in?

metthal: The commented out code is there just for the sake of completeness. Unfortunately, such grammar is not LALR(1) grammar because it contains reduce/reduce conflicts. I kept it there because what I plan to do is to release yaramod 4.x.x and put into README that yaramod 4.x.x is switching to YARA-X compatibility mode and that 3.x.x should be used to maintain the old YARA compatibility. I want to end this message with basically stating that yaramod is about the be completely deprecated by new YARA-X and that further support for 4.x.x is not planned. So 4.x.x is just a version that is about to be phased out of existence and this old code doesn't really bother me.

MatejKastak: Cool, thanks for the explanation :pray:

MatejKastak (approved):

PR #265, opened 2024-09-24
feat: Added support for 'with' expression

metthal:


MatejKastak (approved):
LGTM :+1: but the tests are failing and it seems that the old user defined arrays are broken now, is that expected?

TommYDeeee (approved):
from a grammar point of view it LGTM

TommYDeeee:
> LGTM 👍 but the tests are failing and it seems that the old user defined arrays are broken now, is that expected? (is this prereq #264?)

It looks like some Windows-specific error as other builds are not failing. This also supports it: https://stackoverflow.com/questions/13157671/seh-exception-with-code-0xc0000005-thrown-in-the-test-body
weird.

"I have encountered SEH-exceptions when accessing memory that does not belong to the current process so I recommend looking at memory-related instructions in the constructor and destructor of CSig."

TommYDeeee:
> > LGTM 👍 but the tests are failing and it seems that the old user defined arrays are broken now, is that expected? (is this prereq #264?)
> 
> It looks like some Windows-specific error as other builds are not failing. This also supports it: https://stackoverflow.com/questions/13157671/seh-exception-with-code-0xc0000005-thrown-in-the-test-body weird.
> 
> "I have encountered SEH-exceptions when accessing memory that does not belong to the current process so I recommend looking at memory-related instructions in the constructor and destructor of CSig."

Oh as I look at it more, just `3.12 Debug` reports those issues, Release builds only report:

`unknown file: error: C++ exception with description "Error at 6.20: Syntax error: Unexpected hex string [, expected one of (, them" thrown in the test body.` 
and this just looks like the previous PR is a prerequisite for this PR, so probably all good

MatejKastak (approved):

PR #266, opened 2024-10-26
fix: Expression array type is now properly recognized

metthal:
We were incorrectly marking certain expression arrays like `(true, $str)` as string arrays while they should have been mixed expression arrays.

### DIFF
--- a/include/yaramod/parser/parser_driver.h
+++ b/include/yaramod/parser/parser_driver.h
@@ -54,6 +54,13 @@ enum IncludeResult
 	Included ///< File was successfully included
 };
 
+enum class ExpressionArrayType
+{
+	Undetermined, // If the array is just ($s0, $s1) - it's undetermined but we turn it into string set in the end
+	MixedArray, // For arrays like ($s0, true)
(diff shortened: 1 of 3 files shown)


include/yaramod/parser/parser_driver.h
```diff
@@ -248,7 +260,7 @@ class ParserDriver
 
 	std::vector<std::string> _deferredIncludes; ///< Paths of included files when deferred includes are turned on.
 
-	std::vector<bool> _expressionArrayStack; ///< Indicates that expression array (..., ...) represents string set
+	std::vector<ExpressionArrayType> _expressionArrayStack; ///< Indicates that expression array (..., ...) represents string set
```
MatejKastak, line 263: Nit: I think that the doc string was not updated

MatejKastak (approved):

include/yaramod/parser/parser_driver.h
metthal: Thanks for this. I kind of realized thanks to this that I also forgot to test that the nesting actually works so I've also added test for nesting.

PR #268, opened 2024-12-01
feat: Added possiblity to clone expressions (both AST node + tokenstream)

metthal:
This will make it easier to create copies out of existing expressions without having to rebuild them yourself manually (especially if their structure is not fixed).


vojone (approved):
LGTM. Hope I did not miss anything.

PR #292, opened 2026-06-04
feat: `.len()` method on array, dictionary and string

MichaelTheSynthCat:


include/yaramod/types/symbols.h
```diff
@@ -197,6 +201,20 @@ class FunctionSymbol : public Symbol
 	std::vector<std::vector<std::string>> _overloadArgumentsNames; ///< Names of arguments of all known overloads
 };
 
+inline std::optional<std::shared_ptr<Symbol>> ValueSymbol::getAttribute(const std::string& name) const
```
metthal, line 204: Why does `ValueSymbol` need this method too? 🤔

src/parser/parser_driver.cpp
```diff
@@ -2089,26 +2090,44 @@ void ParserDriver::defineGrammar()
...
 				}
 			}
+			else if (parentSymbol->isArray() || parentSymbol->isDictionary())
+			{
+				auto iterParentSymbol = std::static_pointer_cast<IterableSymbol>(parentSymbol);
+				attr = iterParentSymbol->getAttribute(symbol_token->getString());
+				if (!attr)
+					error_handle(args[2].getTokenIt()->getLocation(), "Unrecognized identifier '" + symbol_token->getString() + "' referenced on array/dictionary");
+			}
+			else if (parentSymbol->isValue())
```
metthal: I'm not sure if this is really needed or is it?

tests/cpp/parser_tests.cpp
```diff
@@ -9068,5 +9068,117 @@ rule formatting_in_expression_arrays_not_broken
...
+)");
+
+	EXPECT_TRUE(driver.parse(input));
+	ASSERT_EQ(1u, driver.getParsedFile().getRules().size());
+
+	const auto& rule = driver.getParsedFile().getRules()[0];
+	EXPECT_EQ("dotnet.user_strings[0].len() >= 3", rule->getCondition()->getText());
+
+	EXPECT_EQ(input_text, driver.getParsedFile().getTextFormatted());
+}
```
MatejKastak, line 9181: Is there any support not only for parsing but also constructing this using the builder pattern?

I was expecting seeing some tests that exercise this. E.g. to programmatically build the rules in the tests above.

MatejKastak (commented):
I would also suggest to also test the Visitors.

include/yaramod/types/symbols.h
MichaelTheSynthCat: Even Strings (represented by `ValueSymbol`) have [`.len()` method](https://virustotal.github.io/yara-x/docs/writing_rules/differences-with-yara/#built-in-len-method).

metthal: Oh that makes sense in that case. But wouldn't it probably make sense to have `getAttribute` on global `Symbol` level then and override it for classes where it makes sense?

MichaelTheSynthCat: I was thinking about that because it would simplify the code a lot. I was just not sure if we would be okay with adding a new virtual method on `Symbol`.

tests/cpp/parser_tests.cpp
MichaelTheSynthCat: Tests added for both Builder and Visitor patterns.

src/parser/parser_driver.cpp
MichaelTheSynthCat: That was intended for Strings too. Right now it has benn removed thanks to the new common `Symbol::getAttribute()` method.

MatejKastak (approved):
Thx for the tests

